PR #1, opened 2017-10-31
fix(resources): crash on datetime.replace

H--o-l:
`datetime.replace()` function doesn't manage days from us.

As consequence `(2017, 10, 31).replace(month=11)` will try to create
the date `(2017, 11, 31)` which doesn't exist.

In such case datetime lib crash with the following error:
```
ValueError: day is out of range for month
```

This commit replace the use of `datetime.replace()` to increment month,
by the use of `datetime + relativedelta()`.

`relativedelta()` function is part of `dateutils`, so I also add the
dependency in `setup.py`.

### DIFF
--- a/stripe_mock_server/resources.py
+++ b/stripe_mock_server/resources.py
@@ -20,6 +20,8 @@ import random
 import string
 import time
 
+from dateutil.relativedelta import relativedelta
+
 from .errors import UserError
 
 
@@ -854,15 +856,9 @@ class Subscription(StripeObject):
         elif plan.interval == 'week':
             current_period_end += timedelta(days=7)
         elif plan.interval == 'month':
-            current_period_end = current_period_start.replace(
-                month=(current_period_start.month % 12) + 1,
-                year=current_period_start.year +
-                int(current_period_start.month / 12))
+            current_period_end += relativedelta(months=1)
         elif plan.interval == 'year':
-            current_period_end = current_period_start.replace(
-                year=current_period_start.year + 1)
-        self.current_period_start = int(current_period_start.timestamp())
-        self.current_period_end = int(current_period_end.timestamp())
+            current_period_end += relativedelta(years=1)
 
         self.items = List('/v1/subscription_items?subscription=' + self.id)
         self.items._list.append(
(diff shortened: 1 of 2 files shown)


stripe_mock_server/resources.py
```diff
@@ -854,10 +856,8 @@ def _set_up_subscription_and_invoice(self, plan):
         elif plan.interval == 'week':
             current_period_end += timedelta(days=7)
         elif plan.interval == 'month':
-            current_period_end = current_period_start.replace(
-                month=(current_period_start.month % 12) + 1,
-                year=current_period_start.year +
-                int(current_period_start.month / 12))
+            delta = relativedelta(months=1)
+            current_period_end = current_period_start + delta
```
adrienverge: While you're at it, can you make it looks nice and consistent for months and years?
```py
        if plan.interval == 'day':
            current_period_end += timedelta(days=1)
        elif plan.interval == 'week':
            current_period_end += timedelta(days=7)
        elif plan.interval == 'month':
            current_period_end += relativedelta(months=1)
        elif plan.interval == 'year':
            current_period_end += relativedelta(years=1)
```

adrienverge:
Also you can rebase on `master`, I just fixed CI tests.

adrienverge:
By the way the same problem would happen for yearly subscriptions, if you're on February 29th and you add one year...

H--o-l:
> By the way the same problem would happen for yearly subscriptions, if you're on February 29th and you add one year...

Yes you are right, I did the PR a bit quick, focussing only on the error we had on the quetzal CI.

> Also you can rebase on master, I just fixed CI tests.

Nice !

> dateutils is an old package that hasn't been updated since 2013 and does not provide what you use in this PR. Did you mean python-dateutil >=2.6.1?
Be careful with that! All those packages in setup.py will be installed when stripe_mock_server is installed.

I really don't like python lib management. Libs doesn't have same names in package manager and when we use them. I guess I will have to get used to it but it's not the first time it confuse me !
Anyway thank for catching it !

---
I fixed the PR and rebase it, should be good !

adrienverge:
> I did the PR a bit quick, focussing only on the error we had on the quetzal CI.

Sure, I understand!

Thanks for fixing that, I'll release a new version.

PR #2, opened 2017-11-02
fix(resources): fix previous fix

H--o-l:
Commit 242e6bdfbaaf00c89ee64ff580c1dda691c6fae1, which aimed at fixing
a crash, removed some lines it shouldn't have.

Let's revert thoses two lines deletion.


adrienverge:
"fix previous fix", sounds poetic

PR #8, opened 2018-03-20
Add Dockerfile and Makefile.

kerak19:
I couldn't use your app, because sudo was needed, so i've containerized it with Docker.
I've created docker image for your app using python:alpine base image.
You should change Makefile to fit your docker ID if you want to update and mantain this image.
I've also updated readme to show how to use docker image.

Change host to 0.0.0.0
Update Readme.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:alpine
+
+ADD . /src
+
+WORKDIR /src
+
+RUN python setup.py install
+
+EXPOSE 8420
+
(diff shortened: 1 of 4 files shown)


adrienverge:
Sorry but I don't see the need for a Docker container in this repo, especially if we lose IPv6 capability.

Why do you need sudo? You can use `pip3 install --user -U localstripe` that doesn't need elevated rights.

kerak19:
Thanks, didn't read pip3 manual yet.

I'm closing PR then.

PR #25, opened 2018-12-05
Add a few more events.

M0dM:
Hi @adrienverge,

Do you think this is correct ?
To be 100% honest I am not sure I've understood all your code.

As I need those events to start using this project, I tried to add them myself...

Any review would be appreciated.


adrienverge:
Hi @M0dM,

Thanks for contributing!

These look like three indenpendant changes, can you make three commits? The `plan.created` and `product.created` parts look perfect (do you confirm you receive what you expect on your endpoint?). I have to test the last one.

Also, since you improve localstripe, can you update the README (there is a list of supported events)?

M0dM:
Close this one to create separate ones.

PR #27, opened 2018-12-06
Add /_config/store route to flush data from store.

M0dM:
Introducing new /_config/store route.
Requesting with method DELETE on it would flush all store data.

This is very useful for me to remove all localstripe data between two integration test.

Edit: squashed

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -209,6 +209,19 @@ Only those events types are currently supported:
 - Subscription: ``customer.subscription.created`` and ``customer.subscription.deleted``
 - Invoice: ``invoice.created``, ``invoice.payment_succeeded`` and ``invoice.payment_failed``
 
+Flush stored data
+-----------------
+
+Flushing data programmatically can be useful to reset localstripe if your are
+using it with any test framework.
+
+Flushing stored data can be performed using the ``/_config/data`` route
+with DELETE http method:
+
+.. code:: shell
+
+ curl -X DELETE localhost:8420/_config/data
+
 Hacking and contributing
 ------------------------
 
--- a/localstripe/server.py
+++ b/localstripe/server.py
@@ -282,7 +282,13 @@ async def config_webhook(request):
     return web.Response()
 
 
+async def flush_store(request):
+    store.clear()
+    return web.Response()
+
+
 app.router.add_post('/_config/webhooks/{id}', config_webhook)
+app.router.add_delete('/_config/data', flush_store)
 
 
 def start():


README.rst
```diff
@@ -205,6 +205,17 @@ currently supported (these include ``customer.created``, ``customer.updated``,
 ``customer.subscription.created``, ``customer.subscription.deleted``
 ``invoice.created``, ``invoice.payment_succeeded``, ``invoice.payment_failed``).
 
+Flush stored data
+------------
+Flushing data programmatically can be useful
+if your are using localstripe with any test framework.
```
adrienverge: A suggestion to be even more explicit:

```
Flushing data programmatically can be useful to reset localstripe if your are
using it with any test framework.
```

README.rst
```diff
@@ -205,6 +205,17 @@ currently supported (these include ``customer.created``, ``customer.updated``,
...
+
+Flushing stored data can be performed using the ``/_config`` route with DELETE http method:
```
adrienverge: You mean ``/_config/store`` or ``/_config/reset``?

M0dM: /_config/store resource and DELETE request seems for me a good idea initially.

But because this project aim to be used for testing only, and "store" is more an internal keyword than something  straightly understandable by any new user I think the best one could of those routes:

* /_config/data + DELETE
* /_data + DELETE

What do you think about this ?

README.rst
M0dM: Seems perfect.

M0dM:
I have updated the doc and the url to /_config/data + DELETE http method.
Don't hesitate to comment if you don't think we've done on this one.

README.rst
```diff
@@ -205,6 +205,18 @@ currently supported (these include ``customer.created``, ``customer.updated``,
 ``customer.subscription.created``, ``customer.subscription.deleted``
 ``invoice.created``, ``invoice.payment_succeeded``, ``invoice.payment_failed``).
 
+Flush stored data
+------------
```
adrienverge: Again, this is invalid RST format.

M0dM: Sure I'll do it.
I was waiting for your review before doing anything on this pull request. ;)

M0dM:
It should be all right now.

adrienverge:
Looks good, just the commit message to fix (it contains extra info that is obsolete).
For info people use Git history as a changelog, it's also very useful for `git bisect`, `git blame` and co.

M0dM:
I tried to follow your commit message convention.
By the way, it could be great to hove some rules about this convention inside the "Hacking and contributing" section of the doc.

adrienverge:
> By the way, it could be great to hove some rules about this convention inside the "Hacking and contributing" section of the doc.

Sure, I agree the contribution conventions should be made clearer -- maybe via a new `CONTRIBUTING.rst` file. But these are kind of standard (at least for having atomic commits and commit messages that describe what the commit does).

Anyway, thanks for this new feature!

PR #30, opened 2018-12-06
Allow retrieving Event stripe objects.

M0dM:
Allow stripe events to be retrieved.

Edit: Squashed

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -549,7 +549,15 @@ class Event(StripeObject):
 
     @classmethod
     def _api_create(cls, **data):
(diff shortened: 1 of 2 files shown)


localstripe/resources.py
```diff
@@ -549,7 +549,15 @@ def __init__(self, type, data):
 
     @classmethod
     def _api_create(cls, **data):
-        raise UserError(400, 'Bad request')
+        raise UserError(405, 'Method Not Allowed')
```
adrienverge: Clearly better than `Bad request` :ok_hand:

PR #41, opened 2018-12-10
Retrieve single user source by nested url.

M0dM:


localstripe/resources.py
```diff
@@ -494,6 +494,29 @@ def _api_delete(cls, id):
...
+        if type(source_id) is str and source_id.startswith('src_'):
+            source_obj = Source._api_retrieve(source_id)
+        elif type(source_id) is str and source_id.startswith('tok_'):
+            source_obj = Token._api_retrieve(source_id).card
+        elif type(source_id) is str and source_id.startswith('card_'):
+            source_obj = Card._api_retrieve(source_id)
+
+        try:
+            assert source_obj is not None
+            assert source_obj.customer == id
```
adrienverge: `Source` objects don't have a `customer` attribute, so the following code produces a crash (error 500):

```bash
HOST=http://localhost:8420
SK=sk_test_12345

sepa_cus=$(
  curl -sSf -u $SK: $HOST/v1/customers \
       -d description='I pay with SEPA debit' \
       -d email=sepa@euro.fr \
  | grep -oE 'cus_\w+' | head -n 1)

src=$(curl -sSf -u $SK: $HOST/v1/sources \
           -d type=sepa_debit \
           -d sepa_debit[iban]=DE89370400440532013000 \
           -d currency=eur \
           -d owner[name]='Jenny Rosen' \
      | grep -oE 'src_\w+')

curl -sSf -u $SK: $HOST/v1/customers/$sepa_cus/sources \
     -d source=$src

# Get a customer source directly:
curl -sSf -u $SK: $HOST/v1/customers/$sepa_cus/sources/$src
```

```
  File ".../localstripe/localstripe/resources.py", line 523, in _api_retrieve_source
    assert source_obj.customer == id
AttributeError: 'Source' object has no attribute 'customer'
```

localstripe/resources.py
```diff
@@ -494,6 +494,29 @@ def _api_delete(cls, id):
...
         return super()._api_delete(id)
 
+    @classmethod
+    def _api_retrieve_source(cls, id, source_id, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        source_obj = None
+
+        # return 404 if does not exist
```
adrienverge: What if customer `id` does not exist? It must also be checked.

localstripe/resources.py
```diff
@@ -494,6 +494,29 @@ def _api_delete(cls, id):
...
+        if type(source_id) is str and source_id.startswith('src_'):
+            source_obj = Source._api_retrieve(source_id)
+        elif type(source_id) is str and source_id.startswith('tok_'):
+            source_obj = Token._api_retrieve(source_id).card
```
adrienverge: Tokens can be associated to customers sources? It seems strange. Usually they're just used to create a new source.

localstripe/resources.py
```diff
@@ -494,6 +494,29 @@ def _api_delete(cls, id):
...
+        elif type(source_id) is str and source_id.startswith('card_'):
+            source_obj = Card._api_retrieve(source_id)
+
+        try:
+            assert source_obj is not None
+            assert source_obj.customer == id
+        except AssertionError:
+            raise UserError(400, 'Bad request')
```
adrienverge: If the source does not exists, `404` should be returned.

Issue #44, opened 2018-12-10
Stripe Plan Creation with Tiers


nrobates:
Trying to use this to create a Plan. However this is the error I keep getting:
`Stripe\Error\InvalidRequest: Unexpected billing_scheme, tiers_mode, usage_type, tiers`

Am I missing something here. I thought this was following the latest version of Stripe API. What version is this supporting? 

Many thanks in advance. Would love to use this package for sure, however need this part to work in order to even continue.

adrienverge:
Hi,

This is trying to follow the latest version of the API, yes, but not all features are implemented yet.

If these are simple options, you can try to contribute. For example here are recent contributions to add options for plans: https://github.com/adrienverge/localstripe/pull/29 https://github.com/adrienverge/localstripe/pull/35.

nrobates:
@marotpam / @adrienverge Just tried out the latest docker image which I expected contained the support for tiered pricing. However, if I use the same example for testing provided or the following, there is an invalid request error returned. Here is an example:

```
curl localhost:8420/v1/plans -u sk_test_12345: \
   -d id=annual-tiered \
   -d name='Annual tiered' \
   -d currency=eur \
   -d interval=year \
   -d interval_count=1 \
   -d usage_type=licensed \
   -d billing_scheme=tiered \
   -d tiers_mode=volume \
   -d tiers[0][unit_amount]=1500 \
   -d tiers[0][up_to]=10 \
   -d tiers[1][unit_amount]=1000 \
   -d tiers[1][up_to]=inf \
```

The response is:
```
{
  "error": {
    "message": "Unexpected billing_scheme, tiers_mode, tiers",
    "type": "invalid_request_error"
  }
}
```

Just wondering if either the tiered support did not make it in, or if there is a bug still. Many Thanks.

adrienverge:
@nrobates thanks for reporting that. I guess I pushed a wrong version to dockerhub. Could you try again now?

nrobates:
@adrienverge yes that fixed the error I was receiving. However there is a few other things that I noticed. 

1. When creating a Plan, Stripe asks for `nickname` not `name`.
2. It would be great to not have to pass `id` along and it could auto-generate by slugging the nickname or something, it is something that can be worked around.

Thanks!

marotpam:
Hi @nrobates, glad the tiered pricing is working.

[...]

PR #51, opened 2019-02-15
Add support for multiple plans and tiered billing in invoices

marotpam:
This should fix the issues @nrobates is describing in https://github.com/adrienverge/localstripe/issues/44 where subscriptions to plans with tiered billing failed because of how the invoices where expecting a plan amount to be given (which is not there for tiered plans). In this PR I also added support for having multiple subscriptions to the same plan, as otherwise plan quantities were always treated as 1, which limited the benefits of using tiered pricing.
Still pending are changes to allow subscribing to different plans in the same subscription, and making the upcoming invoice endpoint work with tiered billing. 
The latter will probably require a refactor to reuse the logic used to calculate the amounts introducing a `Tier` class. That is likely to take some time, so in order to make what we currently have work for @nrobates we can leave that as a future issue

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1096,6 +1096,7 @@ class Plan(StripeObject):
         self.usage_type = usage_type
         self.billing_scheme = billing_scheme
         self.tiers = tiers
+        self.tiers_mode = tiers_mode
 
         schedule_webhook(Event('plan.created', self))
 
@@ -1308,7 +1309,7 @@ class Subscription(StripeObject):
         self.canceled_at = None
         self.discount = None
         self.ended_at = None
-        self.quantity = 1
+        self.quantity = items[0]['quantity']
         self.status = 'active'
         self.trial_end = None
         self.trial_start = None
@@ -1338,7 +1339,10 @@ class Subscription(StripeObject):
 
         self.items = List('/v1/subscription_items?subscription=' + self.id)
         self.items._list.append(
-            SubscriptionItem(subscription=self.id, plan=plan.id, quantity=1))
+            SubscriptionItem(
+                subscription=self.id,
+                plan=plan.id,
+                quantity=self.quantity))
 
         invoice_items = []
 
@@ -1359,7 +1363,7 @@ class Subscription(StripeObject):
             for si in self.items._list:
                 invoice_items.append(
                     InvoiceItem(subscription=self.id, plan=si.plan.id,
-                                amount=si.plan.amount,
+                                amount=si._calculate_amount(),
                                 currency=si.plan.currency,
                                 description=si.plan.name,
                                 customer=self.customer,
@@ -1396,12 +1400,15 @@ class Subscription(StripeObject):
                 for item in items:
                     id = item.get('id', None)
                     plan = item.get('plan', None)
-                    quantity = try_convert_to_int(item.get('quantity', None))
                     assert id is not None or plan is not None
                     if id is not None:
                         assert type(id) is str and id.startswith('si_')
-                    if quantity is not None:
-                        assert type(quantity) is int and quantity > 0
+                    if item.get('quantity', None) is not None:
+                        item['quantity'] = try_convert_to_int(item['quantity'])
+                        assert type(item['quantity']) is int
+                        assert item['quantity'] > 0
+                    else:
+                        item['quantity'] = 1
         except AssertionError:
             raise UserError(400, 'Bad request')
 
@@ -1410,6 +1417,8 @@ class Subscription(StripeObject):
 
         if items is None or len(items) != 1 or not items[0]['plan']:
             raise UserError(500, 'Not implemented')
+
+        self.quantity = items[0]['quantity']
 
         plan = Plan._api_retrieve(items[0]['plan'])
         # If the subscription is updated to a more expensive plan, an invoice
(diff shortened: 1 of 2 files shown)


localstripe/resources.py
```diff
@@ -1308,7 +1309,7 @@ def __init__(self, customer=None, metadata=None, items=None,
         self.canceled_at = None
         self.discount = None
         self.ended_at = None
-        self.quantity = 1
+        self.quantity = items[0]['quantity']
```
marotpam: This is what will need to change when we support plans with different subscriptions. For now I'm leaving it like this as nothing else is prepared to handle that scenario

localstripe/resources.py
```diff
@@ -1471,6 +1476,45 @@ def __init__(self, subscription=None, plan=None, quantity=1,
...
+                    quantity -= int(t['up_to'])
+
+            return amount
+
+        return 0
+
+    def _calculate_amount_in_tier(self, quantity, index):
+        t = self.plan.tiers[index]
+        return int(t["unit_amount"])*quantity + int(t["flat_amount"])
+
```
marotpam: This could be simplified if with an extra `Tier` class that could be extended to cover the different cases, but will do for now. @adrienverge please if you see something that is not very idiomatic I'm let me know so I can change it :)

localstripe/resources.py
```diff
@@ -1415,7 +1420,7 @@ def _update(self, metadata=None, items=None, tax_percent=None,
         # If the subscription is updated to a more expensive plan, an invoice
         # is not automatically generated. To achieve that, an invoice has to
         # be manually created using the POST /invoices route.
-        self._set_up_subscription_and_invoice(plan, create_an_invoice=False)
+        self._set_up_subscription_and_invoice(plan, quantity)
```
adrienverge: - Why do you remove `create_an_invoice=`? It is not consistent with the comment above.
- `quantity` is not defined if `items` is `None` or `[]`, which will lead to a crash.
- Do you mean that an invoice should be created only if `quantity` ≠ 0? In this case, maybe this would be clearer: `create_an_invoice = quantity > 0; _set_up_subscription_and_invoice(plan, create_an_invoice=create_an_invoice)`?
  If it's not what you mean, can you explain?

localstripe/resources.py
adrienverge: You're right! But I'd rather not create a `Tier` class: the localstripe API relies on the fact that `StripeObject`s are gettable, listable, updatable... which is not the case with tiers.

test.sh
```diff
@@ -210,3 +231,33 @@ curl -sSf -u $SK: $HOST/v1/invoices/upcoming?customer=$cus
...
+
+curl -sSf -u $SK: $HOST/v1/customers/$cus/sources \
+     -d source=$tok
+
+curl -sSf -u $SK: $HOST/v1/subscriptions \
+      -d customer=$cus \
+      -d items[0][plan]=monthly-tiered-graduated \
+      -d items[0][quantity]=5
+
+curl -sSf -u $SK: $HOST/v1/invoices?customer=$cus
```
adrienverge: Please leave the newline at the end of the file (see the icon GitHub shows ;))

localstripe/resources.py
marotpam: You're absolutely right, I shouldn't have removed `create_an_invoice`, good catch! I'll update the PR to fix this

localstripe/resources.py
marotpam: In this case `Tier` wouldn't inherit from `StripeObject`, it would be just a Value Object that could be used to make the code clearer. We can revisit that idea in the future if new logic that varies upon the type of tiers is added, for now it's simple enough

adrienverge:
Thanks for this update! I allowed myself to push to your branch with minor style edits.

We're very close, but currently it crashes my test suite:

      File "~/.local/lib/python3.7/site-packages/localstripe/resources.py", line 1418, in _update
        self.quantity = items[0]['quantity']
    KeyError: 'quantity'

My guess is that the same logic as in https://github.com/adrienverge/localstripe/blob/c157a921ceb72ae06b525d53b89ff45fd6576591/localstripe/resources.py#L1285-L1290 should be applied to `_update()`. Also, can you add a test to make sure the problem won't come back? For instance, something like:
```diff
--- a/test.sh
+++ b/test.sh
@@ -255,9 +255,13 @@ cus=$(curl -sSf -u $SK: $HOST/v1/customers \
 curl -sSf -u $SK: $HOST/v1/customers/$cus/sources \
      -d source=$tok
 
-curl -sSf -u $SK: $HOST/v1/subscriptions \
-      -d customer=$cus \
-      -d items[0][plan]=monthly-tiered-graduated \
-      -d items[0][quantity]=5
+sub=$(curl -sSf -u $SK: $HOST/v1/subscriptions \
+           -d customer=$cus \
+           -d items[0][plan]=monthly-tiered-graduated \
+           -d items[0][quantity]=5 \
+      | grep -oE 'sub_\w+' | head -n 1)
+
+curl -sSf -u $SK: $HOST/v1/subscriptions/$sub \
+     -d items[0][plan]=annual-tiered-volume
 
 curl -sSf -u $SK: $HOST/v1/invoices?customer=$cus
```

marotpam:
Hi @adrienverge finally got some time to fix the issue with updating subscriptions, the problem was that quantity in plans was not defaulting to 1 when it was not being set. While testing I also saw there's a comment that is not quite true:
```
# If the subscription is updated to a more expensive plan, an invoice
# is not automatically generated. To achieve that, an invoice has to
# be manually created using the POST /invoices route.
```
because invoices are indeed created and charged in Stripe when updating a subscription results in changing the subscription cycle (i.e. from a monthly to a yearly period). I didn't do any work on that as it seems to be an existing lack of functionality, so we can work on that on a separate PR

Cheers

adrienverge:
@marotpam thanks for the update!

I squashed your commits, will merge and create a new release.

PR #56, opened 2019-06-18
Implement tax rates

adrienverge:
#### fix: Accept booleans passed in x-www-form-urlencoded

Unlike JSON data, boolean values sent as
`application/x-www-form-urlencoded` are represented as strings. In this
example `active` has the value `"true"`:

    curl $HOST/v1/products -d name=T-shirt -d active=true

Let's decode `"false"` and `"true"` when needed.

---

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1616,6 +1616,42 @@ class TaxId(StripeObject):
         self.verification = {'status': 'pending'}
 
 
+class TaxRate(StripeObject):
+    object = 'tax_rate'
+    _id_prefix = 'txr_'
+
+    def __init__(self, display_name=None, inclusive=None, percentage=None,
+                 active=True, description=None, jurisdiction=None,
+                 metadata=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        inclusive = try_convert_to_bool(inclusive)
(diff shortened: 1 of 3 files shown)


localstripe/resources.py
```diff
@@ -1616,6 +1616,42 @@ def __init__(self, country=None, customer=None, type=None, value=None,
...
+            assert type(inclusive) is bool
+            assert type(percentage) is float
+            assert type(active) is bool
+            assert percentage >= 0 and percentage <= 100
+            assert description is None or type(description) is str
+            assert jurisdiction is None or type(jurisdiction) is str
+        except AssertionError:
+            raise UserError(400, 'Bad request')
+
+        # All exceptions must be raised before this point.
```
H--o-l: Oh? 
Is it a convention, or it have unexpected side effects?

adrienverge: Run `grep 'All exceptions must be raised before this point.' localstripe/resources.py`, you'll see that this is used in every class 😉

Put briefly, the goal is raise user errors (HTTP 400) before starting any non-reversable work (create the object itself in the *store*, create a sub-resource, trigger the creation of a charge and invoice because the subscription is modifies, trigger webhooks...). In this case (`class TaxRate`), the only non-reversable work is done in `super().__init__()`: this definitely creates an unique ID and adds the object in the store (i.e. it will be list in API get all).

PR #85, opened 2019-08-18
Allow setting product.id

mikhailmoshnogorsky:
and small fixes

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1621,7 +1621,7 @@ class Product(StripeObject):
     object = 'product'
     _id_prefix = 'prod_'
 
-    def __init__(self, name=None, type=None, active=True, caption=None,
+    def __init__(self, id=None, name=None, type=None, active=True, caption=None,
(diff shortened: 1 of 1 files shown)


adrienverge:
Hi @mikhailmoshnogorsky, thanks for contributing!

About product: 
- You need to check that `id` is not a boolean, an array, an empty string or whatever, see *plans* for an example:
  https://github.com/adrienverge/localstripe/blob/f5cd4f3948553d11e27ee20bd4a80f5f87fe2bed/localstripe/resources.py#L1558
- Can you add a test, maybe just a new `curl` before https://github.com/adrienverge/localstripe/blob/f5cd4f3948553d11e27ee20bd4a80f5f87fe2bed/test.sh#L130, to make sure we now can pass `-d id=productid`?

About tiers, maybe you want to open another PR for them? The change looks good and can be merged (although CI tests are failing, I'm wondering what results on Travis say).

adrienverge:
Obsoleted by https://github.com/adrienverge/localstripe/pull/103.

PR #90, opened 2019-08-30
add support for pre auth charges (charge capture param)

grelca:
accept the capture parameter in charge creation, and create `POST /v1/charges/:id/capture`


adrienverge:
Looks very nice!

Would that be easy to add a very simple test case in `test.sh`, to make sure there will be no regressions in the future?

PR #91, opened 2019-08-31
bugfix: only parse charge source as a string if it is one

grelca:
Customer._get_default_payment_method_or_source() returns a Card/Source/PaymentMethod object, not a string, so when only a customer is provided, you get an error like:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/aiohttp/web_protocol.py", line 418, in start
    resp = await task
  File "/usr/local/lib/python3.7/site-packages/aiohttp/web_app.py", line 458, in _handle
    resp = await handler(request)
  File "/usr/local/lib/python3.7/site-packages/aiohttp/web_middlewares.py", line 119, in impl
    return await handler(request)
  File "/root/localstripe/localstripe/server.py", line 56, in error_middleware
    return await handler(request)
  File "/root/localstripe/localstripe/server.py", line 192, in auth_middleware
    return await handler(request)
  File "/root/localstripe/localstripe/server.py", line 204, in f
    return json_response(cls._api_create(**data)._export(expand=expand))
  File "/root/localstripe/localstripe/resources.py", line 133, in _api_create
    return cls(**data)
  File "/root/localstripe/localstripe/resources.py", line 346, in __init__
    if source.startswith('pm_'):
AttributeError: 'Card' object has no attribute 'startswith'
```

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -339,8 +339,7 @@ class Charge(StripeObject):
             source = customer_obj._get_default_payment_method_or_source()
             if source is None:
                 raise UserError(404, 'This customer has no payment method')
-
-        if source.startswith('pm_'):
+        elif source.startswith('pm_'):
             source = PaymentMethod._api_retrieve(source)
         elif source.startswith('src_'):
             source = Source._api_retrieve(source)
--- a/test.sh
+++ b/test.sh
@@ -246,6 +246,12 @@ ds=$(curl -sSf -u $SK: $HOST/v1/customers/$cus?expand%5B%5D=default_source \
      | grep -oE '"default_source": null",' || true)
 [ -z "$ds" ]
 
+# we can charge a customer without specifying the source
+curl -sSf -u $SK: $HOST/v1/charges \
+     -d customer=$cus \
+     -d amount=1000 \
+     -d currency=usd
+
 curl -sSf -u $SK: $HOST/v1/invoices?customer=$cus
 
 code=$(curl -s -o /dev/null -w "%{http_code}" -u $SK: \


adrienverge:
Hi @grelca, thanks for detecting and fixing this bug!

Just an idea: since `source` is already checked to be a string (a few lines above), I think the following would make more sense:

```python
        if source is None:
            customer_obj = Customer._api_retrieve(customer)
            source = customer_obj._get_default_payment_method_or_source()
            if source is None:
                raise UserError(404, 'This customer has no payment method')
        elif source.startswith('pm_'):
            source = PaymentMethod._api_retrieve(source)
        elif source.startswith('src_'):
            source = Source._api_retrieve(source)
        elif source.startswith('card_'):
            source = Card._api_retrieve(source)
        else:
```
(and it would avoid bugs if the user provides something like `source: 'unknown_4Kb5Ac46'`)

grelca:
well, in the assertions it doesnm't only check that it's a string - it does check that it's a string that starts with `pm_`, `card_` or `src_` so that should already be a protected case.

the reason i put that into a second `if` statement rather than making `elif`s is because i think (and i could absolutely be wrong here, i did not dig into the code that would go down this path) that `_get_default_payment_method_or_source` _can_ return a string, if the customer has invoice settings. again i didn't really dig into what this means, but i thought this first path had the potential to return a string, so if it did i wanted it to still go through the second lookup portion.

```python
    def _get_default_payment_method_or_source(self):
        if self.invoice_settings.get('default_payment_method'):
            return PaymentMethod._api_retrieve(
                self.invoice_settings['default_payment_method'])
        elif self.default_source:
            return [s for s in self.sources._list
                    if s.id == self.default_source][0]
```

grelca:
...and i just realized that the invoice settings part uses a `PaymentMethod._api_retrieve` so yeah i was definitely wrong lol

PR #94, opened 2019-09-04
support for update + delete source

grelca:
since updating and deleting cards/sources does not happen on the basic `/v1/sources/{id}` path, adds paths to support them at `v1/customers/{customer_id}/sources/{source_id}` that hook into the main `Source._api_delete` and `Source._api_update`

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -631,6 +631,11 @@ class Customer(StripeObject):
         return source_obj
 
     @classmethod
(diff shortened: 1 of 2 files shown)


adrienverge:
Looks good. GitHub says there are conflicts: can you rebase and squash all commits into one?

grelca:
@adrienverge sure thing! done :)

PR #96, opened 2019-09-05
misc(docker) add alpine-based Dockerfile

abreksa4:


adrienverge:
OK to use the alpine-based image, however I prefer not having/maintaining the `Dockerfile` (or any package source file) in the repository. Can you update the snippet in `README.rst` instead?

PR #97, opened 2019-09-05
preserve source fingerprint across times added

grelca:
a source fingerprint is supposed to always remain the same for the same source. this makes it so :)


adrienverge:
Perfect. Thanks!

PR #100, opened 2019-09-12
resource field updates

dlsteuer:
- allow the name field for customers
- add trial_period_days field
- when on a free trial don't invoice (specified by trial_end or trial_period_days)

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -2023,18 +2023,22 @@ class Subscription(StripeObject):
     def __init__(self, customer=None, metadata=None, items=None,
                  tax_percent=None,  # deprecated
                  enable_incomplete_payments=True,  # legacy support
-                 payment_behavior='allow_incomplete', **kwargs):
+                 payment_behavior='allow_incomplete',
+                 trial_period_days=None, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
         tax_percent = try_convert_to_float(tax_percent)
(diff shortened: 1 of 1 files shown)


localstripe/resources.py
```diff
@@ -558,6 +558,7 @@ def __init__(self, description=None, email=None, invoice_settings=None,
         self.discount = None
         self.shipping = None
         self.default_source = None
+        self.name = name or ''
```
adrienverge: If `name` is provided, it needs to be checked to be a string (like for `email` and `description` above, line 518).

Also, can you put it before `email` and `description`, for clarity? (both here and in `__init__()` arguments)

localstripe/resources.py
```diff
@@ -2058,12 +2060,15 @@ def __init__(self, customer=None, metadata=None, items=None,
         self.status = 'incomplete'
         self.trial_end = None
         self.trial_start = None
+        self.trial_period_days = trial_period_days
```
adrienverge: If `trial_period_days` is provided, it needs to be checked to be an integer. See an example here for `Charge`:
https://github.com/adrienverge/localstripe/blob/3d0110ac0df320b24ffe275a5ba3645a97eef2e6/localstripe/resources.py#L328

dlsteuer:
Sure thing, changes have been, I think I got the validation the way you asked for.

adrienverge:
I've split your PR into two commits for clarity:

- feat(customers): add support for names in customers
- feat(subscriptions): don't invoice when on a free trial

PR #101, opened 2019-09-18
Bulk compability enhancements with stripe api

### LINKED ISSUE #99: Missing endpoint /customers/:id/subscriptions
It seems that the endpoint `/customers/:id/subscriptions` is not implemented.

Official stripe: `GET https://api.stripe.com/v1/customers/cus_[...]/subscriptions`
```json
{
  "object": "list",
  "data": [],
  "has_more": false,
  "url": "/v1/customers/cus_[...]/subscriptions"
}
```

Localstripe: `GET localhost:8420/v1/customers/cus_[...]/subscriptions`
```
404: Not Found
```

Is there any plans to implement this endpoint? 

That the rest of the integration is working quite well with my test suite. I only ran into the missing support for testing card-tokens (as pointed out in another issue), which was an easy fix 👍 

Cheers

rasmuscnielsen:
This PR fixes #99 + further issues found while running our test suite to against Localstripe.

In highlights

- Adding endpoints for `/customers/:id/subscriptions[/:id]`
- Making the 'items' dictionary optional for `Subscription`. Using the Stripe PHP SDK it sets quantity and plan directly on the model - not in an 'items' dictionary
- Allow to update only 'quantity' on a subscription - without changing plans
- Add missing support for `trial_end`, `cancel_at_period_end`, `prorate`. All behavior may not be 100% in line with Stripe, but at least it does not throw exception

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -782,7 +782,6 @@ extra_apis.extend((
     ('DELETE', '/v1/customers/{id}/sources/{source_id}',
      Customer._api_remove_source),
 
-
     # Subscription endpoints
     ('GET', '/v1/customers/{id}/subscriptions', Customer._api_list_subscriptions),
     ('POST', '/v1/customers/{id}/subscriptions', Customer._api_add_subscription),
@@ -2063,7 +2062,7 @@ class Subscription(StripeObject):
     object = 'subscription'
     _id_prefix = 'sub_'
 
-    def __init__(self, customer=None, metadata=None, items=None,
+    def __init__(self, customer=None, metadata=None, items=None, trial_end=None,
                  plan=None, quantity=None, # optional "items" dictionary
                  tax_percent=None,  # deprecated
                  enable_incomplete_payments=True,  # legacy support
@@ -2075,12 +2074,19 @@ class Subscription(StripeObject):
         if items is None:
             items = [{ 'plan': plan, 'quantity': quantity }]
 
+        if type(trial_end) is str and trial_end == 'now':
+            trial_end = time.time()
+
+        trial_end = try_convert_to_int(trial_end)
         tax_percent = try_convert_to_float(tax_percent)
-        enable_incomplete_payments = try_convert_to_bool(
-            enable_incomplete_payments)
+        enable_incomplete_payments = try_convert_to_bool(enable_incomplete_payments)
         trial_period_days = try_convert_to_int(trial_period_days)
+
         try:
             assert type(customer) is str and customer.startswith('cus_')
+            if trial_end is not None:       
+                assert type(trial_end) is int
+                assert trial_end > 1500000000
             if tax_percent is not None:
                 assert type(tax_percent) is float
                 assert tax_percent >= 0 and tax_percent <= 100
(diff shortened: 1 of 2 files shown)


rasmuscnielsen:
Not sure why the build fails. I'd be happy to fix it provided some guidance :-)

adrienverge:
Hi @rasmuscnielsen, thanks for contributing. Can you split this change into multiple PRs, with proper commit names? It will ease reviewing, and is needed for changelog and debugging with `git blame`/`git revert`.

rasmuscnielsen:
Hi @adrienverge 

I've split it up into multiple, more descriptive, commits now. 

Would you still like to have it as multiple PR's? 
Some of them may depend on each other (not 100% sure if there'd be conflicts), in which case we could only do 1 PR at a time.

Either way, it should be quite clear what each commit does now in the 'Files changed' tab :-) 

Let me know how you'd like to proceed.

localstripe/resources.py
```diff
@@ -736,6 +736,39 @@ def _api_list_tax_ids(cls, id, **kwargs):
...
+
+    @classmethod
+    def _api_retrieve_subscription(cls, id, subscription_id, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        obj = Subscription._api_retrieve(subscription_id)
+
+        if obj.customer != id:
+            raise UserError(400, 'Customer ' + id + ' does not have a subscription with ID ' + subscription_id)
```
adrienverge: Isn't it a 404 instead?

If this is the case, a `UserError(404, 'Not Found')` should be enough (and will suppress the linting error).

localstripe/resources.py
```diff
@@ -748,6 +781,14 @@ def _api_list_tax_ids(cls, id, **kwargs):
     # Delete single source by id:
     ('DELETE', '/v1/customers/{id}/sources/{source_id}',
      Customer._api_remove_source),
+
+
```
adrienverge: Can you please remove this empty lines (and the same below), for consistency with the rest of the file?

localstripe/resources.py
```diff
@@ -2064,13 +2064,17 @@ class Subscription(StripeObject):
     _id_prefix = 'sub_'
 
     def __init__(self, customer=None, metadata=None, items=None,
+                 plan=None, quantity=None, # optional "items" dictionary
```
adrienverge: I don't see these parameters in Stripe official docs: https://stripe.com/docs/api/subscriptions/update

Isn't this an old deprecated way of updating a subscription?

localstripe/resources.py
```diff
@@ -2075,12 +2074,19 @@ def __init__(self, customer=None, metadata=None, items=None,
         if items is None:
             items = [{ 'plan': plan, 'quantity': quantity }]
 
+        if type(trial_end) is str and trial_end == 'now':
```
adrienverge: You don't need the `type(trial_end) is str and` here: `if trial_end == 'now'` is enough.

localstripe/resources.py
rasmuscnielsen: The functionality is supported by the 1st party stripe-php sdk: https://github.com/stripe/stripe-php/blob/0e24d89888e4bf051003c45c447b32e5298df55c/lib/Subscription.php#L36

Simply update the quantity directly on the model and `save()`. 

This is also used by the 1st party Laravel package 'Cashier' which our project uses:

https://github.com/laravel/cashier/blob/39916c390454ff21cf25e28fb517eac6f13bd4bd/src/Subscription.php#L370-L376 

For these reasons I'd say there is no sign of it being deprecated until their own SDK stops supporting it :-)

localstripe/resources.py
rasmuscnielsen: Yeah you're right, it's a 404. I can change that. However I thought it'd be nice to have the same error message as the live api which is why I included the actual message.
Unless you'd still rather just have the generic `UserError(404, 'Not Found')`, though could just as well relate to the customer itself (?)  
image

rasmuscnielsen:
Hi @adrienverge 

Sorry for my late reply, and thanks for taking your time to give feedback. Unfortunately I had to move on to another project so haven't had the chance to implement your feedback. 

I'd be happy to make the easy fixes such as spacing and type-check. The thing is that splitting up the PR while also introducing more changes on the right branch will be a bit more cumbersome and time-consuming I feel. Also, all the changes are necessary for our test-suite to pass, so we're quite dependent on all the changes to be accepted - otherwise we'll have to fork anyway.

Furthermore I'm a bit challenged on how much more time I can invest in contributing these changes back into the project. 
I'd really like to make it happen so other people can enjoy the fixes too and of course we won't have to maintain a fork. But at this time I've already spent a couple days on getting it to work with our project, so I really have to wrap it up ASAP, whatever way possible 😐 

Hope your understand my tricky position

localstripe/resources.py
adrienverge: Having the same as Stripe (404 + your original text) seems perfect :+1:

adrienverge:
Hi @rasmuscnielsen, I understand your position. However I can't do the work myself (like you, I lack time), and I won't accept code without agreeing with it, because I have to maintain it once merged. If you fork, you'll see that it's quite some work to update, release, debug problems, etc.

I still think multiple PRs would be easier and faster, especially because some commits can be merged very quick. I'll do a PR for product IDs.

rasmuscnielsen:
Hey @adrienverge 

Thanks for merging some of the commits. This made it faster to extract the rest to separate branches.

Please checkout 

- https://github.com/adrienverge/localstripe/pull/107
- https://github.com/adrienverge/localstripe/pull/106

These seems to be the remainders to make things green in our end.

Cheers

PR #107, opened 2019-11-20
Allow updating a subscription without a plan id

rasmuscnielsen:
When updating a subscription without sending a plan-id, it will simply stay on the same plan.
This is in accordance with the official stripe api.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -2344,7 +2344,6 @@ class Subscription(StripeObject):
                 assert type(items) is list
                 for item in items:
                     id = item.get('id', None)
-                    plan = item.get('plan', None)
                     if id is not None:
                         assert type(id) is str and id.startswith('si_')
                     if item.get('quantity', None) is not None:
@@ -2360,15 +2359,12 @@ class Subscription(StripeObject):
         except AssertionError:
             raise UserError(400, 'Bad request')
 
-        if tax_percent is not None:
-            self.tax_percent = tax_percent
-
         if items is None or len(items) != 1:
             raise UserError(500, 'Not implemented')
 
         # If no plan specified in update request, we stay on the current one
-        if not items[0]['plan']:
-            items[0]['plan'] = self.items._list[0].plan.id
+        if not items[0].get('plan', None):
+            items[0]['plan'] = self.plan.id
 
         # To return 404 if not existant:
         new_plan = Plan._api_retrieve(items[0]['plan'])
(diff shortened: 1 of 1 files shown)


localstripe/resources.py
```diff
@@ -2345,7 +2345,6 @@ def _update(self, metadata=None, items=None, trial_end=None,
                 for item in items:
                     id = item.get('id', None)
                     plan = item.get('plan', None)
```
adrienverge: This line becomes useless, doesn't it?

localstripe/resources.py
```diff
@@ -2361,9 +2360,16 @@ def _update(self, metadata=None, items=None, trial_end=None,
         except AssertionError:
             raise UserError(400, 'Bad request')
 
-        if items is None or len(items) != 1 or not items[0]['plan']:
+        if tax_percent is not None:
+            self.tax_percent = tax_percent
```
adrienverge: I don't understand this change, why is it?

localstripe/resources.py
```diff
@@ -2361,9 +2360,16 @@ def _update(self, metadata=None, items=None, trial_end=None,
...
+
+        if items is None or len(items) != 1:
             raise UserError(500, 'Not implemented')
 
+        # If no plan specified in update request, we stay on the current one
+        if not items[0]['plan']:
+            items[0]['plan'] = self.items._list[0].plan.id
```
adrienverge:
```suggestion
            items[0]['plan'] = self.plan.id
```
will be more straightforward :+1:

localstripe/resources.py
```diff
@@ -2361,9 +2360,16 @@ def _update(self, metadata=None, items=None, trial_end=None,
...
 
-        if items is None or len(items) != 1 or not items[0]['plan']:
+        if tax_percent is not None:
+            self.tax_percent = tax_percent
+
+        if items is None or len(items) != 1:
             raise UserError(500, 'Not implemented')
 
+        # If no plan specified in update request, we stay on the current one
+        if not items[0]['plan']:
```
adrienverge: Can you use `item.get('plan', None)` to avoid crashing with `IndexError` if `plan` is not provided in`items`?

rasmuscnielsen:
Hi @adrienverge 

Thanks for your comments, I believed I addressed them all, and tests seems to be passing now.

localstripe/resources.py
rasmuscnielsen: Good catch - believe it was a bad conflict resolution when I merged

adrienverge:
Thanks for your contributions @rasmuscnielsen!

I'll release a new version soon. I hope it will be fine for your project.

PR #111, opened 2019-12-13
Stripe JS handleCardSetup and handleCardPayment renaming

H--o-l:
### feat(localstripe.js): Remove `address_zip` from `Element`

From a careful riding of the Stripe documentation, it seems that [`Element`]
does not have a `address_zip` field, but a `value.postal_code` field.
In the meantime, the function [`handleCardSetup()`] that consume it, seems
to look for postal code inside
`data.payment_method.billing_details.address.postal_code`, because, in the
background, it use the [`PaymentMethod`] API.

Let's use this structure, it will easy future compatibility with new
[`confirmCardSetup()`].

User be able to call the API like that which wasn't possible before:
```
Stripe().handleCardSetup(clientSecret, {
  {
    "number": "4242424242424242",
    "exp_month": 12,
    "exp_year": 2020,
    "cvc": "314",
  },
  payment_method_data: {
    billing_details: {
      address: {
        postal_code: '12345',
      },
    },
  },
});
```

[`Element`]: https://stripe.com/docs/js/element
[`handleCardSetup()`]: https://stripe.com/docs/js/deprecated/handle_card_payment_element
[`PaymentMethod`]: https://stripe.com/docs/api/payment_methods/object
[`confirmCardSetup()`]: https://stripe.com/docs/js/setup_intents/confirm_card_setup

---
### feat(localstripe.js): Rename `handleCardSetup` to `confirmCardSetup`

From Stripe documentation:
> handleCardSetup has been renamed to confirmCardSetup. In addition to the
> rename, we have slightly modified the arguments. These changes do not affect
> the behavior of the method. While we have no plans to ever remove support for
> handleCardSetup, we think the new name and arguments are easier to understand
> and better convey what the method is doing.

See:
https://stripe.com/docs/stripe-js/reference#stripe-handle-card-setup
https://stripe.com/docs/stripe-js/reference#stripe-confirm-card-setup

---
### feat(localstripe.js): Rename `handleCardPayment` → `confirmCardPayment`

From Stripe documentation:
> handleCardPayment has been renamed to confirmCardPayment. In addition to the
> rename, we have slightly modified the arguments. These changes do not affect
> the behavior of the method. While we have no plans to ever remove support for
> handleCardPayment, we think the new name and arguments are easier to
> understand and better convey what the method is doing.

See:
https://stripe.com/docs/stripe-js/reference#stripe-handle-card-payment
https://stripe.com/docs/stripe-js/reference#stripe-confirm-card-payment

### DIFF
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -98,29 +98,32 @@ class Element {
       exp_month: null,
       exp_year: null,
       cvc: null,
-      address_zip: null,
+      postal_code: null,
     };
 
     const changed = event => {
       this.value = {
-        number: inputs.number.value,
-        exp_month: inputs.exp_month.value,
-        exp_year: '20' + inputs.exp_year.value,
-        cvc: inputs.cvc.value,
-        address_zip: inputs.address_zip.value,
-      };
+        card: {
+          number: inputs.number.value,
+          exp_month: inputs.exp_month.value,
+          exp_year: '20' + inputs.exp_year.value,
+          cvc: inputs.cvc.value,
+        },
+        postal_code: inputs.postal_code.value,
+      }
 
-      if (event.target === inputs.number && this.value.number.length >= 16) {
+      if (event.target === inputs.number &&
+          this.value.card.number.length >= 16) {
         inputs.exp_month.focus();
       } else if (event.target === inputs.exp_month &&
-                 parseInt(this.value.exp_month) > 1) {
+                 parseInt(this.value.card.exp_month) > 1) {
         inputs.exp_year.focus();
       } else if (event.target === inputs.exp_year &&
-                 this.value.exp_year.length >= 4) {
+                 this.value.card.exp_year.length >= 4) {
         inputs.cvc.focus();
       } else if (event.target === inputs.cvc &&
-                 this.value.cvc.length >= 3) {
-        inputs.address_zip.focus();
+                 this.value.card.cvc.length >= 3) {
+        inputs.postal_code.focus();
       }
 
       (this.listeners['change'] || []).forEach(handler => handler());
@@ -131,7 +134,7 @@ class Element {
       inputs[field].setAttribute('type', 'text');
       inputs[field].setAttribute('placeholder', field);
       inputs[field].setAttribute('size', field === 'number' ? 16 :
-                                         field === 'address_zip' ? 5 :
+                                         field === 'postal_code' ? 5 :
                                          field === 'cvc' ? 3 : 2);
       inputs[field].oninput = changed;
       domElement.appendChild(inputs[field]);
@@ -154,11 +157,11 @@ Stripe = (apiKey) => {
         },
       };
     },
-    createToken: async (card) => {
+    createToken: async (element) => {
       console.log('localstripe: Stripe().createToken()');
       let body = [];
-      Object.keys(card.value).forEach(field => {
-        body.push('card[' + field + ']=' + card.value[field]);
+      Object.keys(element.value.card).forEach(field => {
+        body.push('card[' + field + ']=' + card.value.card[field]);
       });
       body.push('key=' + apiKey);
       body.push('payment_user_agent=localstripe');
@@ -217,6 +220,18 @@ Stripe = (apiKey) => {
       try {
         const seti = clientSecret.match(/^(seti_\w+)_secret_/)[1];
         const url = `${LOCALSTRIPE_SOURCE}/v1/setup_intents/${seti}/confirm`;
+        if (element instanceof Element) {
+          data.payment_method_data =
+            data.payment_method_data || {};
+          data.payment_method_data.card = element.value.card;
+          data.payment_method_data.billing_details =
+            data.payment_method_data.billing_details || {};
+          data.payment_method_data.billing_details.address =
+            data.payment_method_data.billing_details.address || {};
+          data.payment_method_data.billing_details.address.postal_code =
+            data.payment_method_data.billing_details.address.postal_code ||
+            element.value.postal_code;
+        }
         let response = await fetch(url, {
           method: 'POST',
           body: JSON.stringify({
@@ -225,15 +240,7 @@ Stripe = (apiKey) => {
             client_secret: clientSecret,
             payment_method_data: {
               type: 'card',
-              card: {
-                number: element.value.number,
-                exp_month: element.value.exp_month,
-                exp_year: element.value.exp_year,
-                cvc: element.value.cvc,
-              },
-              billing_details: {
-                address: {postal_code: element.value.address_zip},
-              },
+              ...data.payment_method_data,
             },
           }),
         });
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -215,21 +215,20 @@ Stripe = (apiKey) => {
     },
     retrieveSource: () => {}, // TODO
 
-    handleCardSetup: async (clientSecret, element, data) => {
-      console.log('localstripe: Stripe().handleCardSetup()');
+    confirmCardSetup: async (clientSecret, data) => {
+      console.log('localstripe: Stripe().confirmCardSetup()');
       try {
         const seti = clientSecret.match(/^(seti_\w+)_secret_/)[1];
         const url = `${LOCALSTRIPE_SOURCE}/v1/setup_intents/${seti}/confirm`;
-        if (element instanceof Element) {
-          data.payment_method_data =
-            data.payment_method_data || {};
-          data.payment_method_data.card = element.value.card;
-          data.payment_method_data.billing_details =
-            data.payment_method_data.billing_details || {};
-          data.payment_method_data.billing_details.address =
-            data.payment_method_data.billing_details.address || {};
-          data.payment_method_data.billing_details.address.postal_code =
-            data.payment_method_data.billing_details.address.postal_code ||
+        if (data.payment_method.card instanceof Element) {
+          const element = data.payment_method.card;
+          data.payment_method.card = element.value.card;
+          data.payment_method.billing_details =
+            data.payment_method.billing_details || {};
+          data.payment_method.billing_details.address =
+            data.payment_method.billing_details.address || {};
+          data.payment_method.billing_details.address.postal_code =
+            data.payment_method.billing_details.address.postal_code ||
             element.value.postal_code;
         }
         let response = await fetch(url, {
@@ -240,7 +239,7 @@ Stripe = (apiKey) => {
             client_secret: clientSecret,
             payment_method_data: {
               type: 'card',
-              ...data.payment_method_data,
+              ...data.payment_method,
             },
           }),
         });
@@ -284,6 +283,14 @@ Stripe = (apiKey) => {
         }
       }
     },
+    handleCardSetup:  // deprecated
+      async function (clientSecret, element, data) {
+        return this.confirmCardSetup(clientSecret, {
+          payment_method: {
+            card: element,
+            ...data.payment_method_data,
+          }});
+      },
     handleCardPayment: async (clientSecret, element, data) => {
       console.log('localstripe: Stripe().handleCardPayment()');
       try {
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -291,8 +291,8 @@ Stripe = (apiKey) => {
             ...data.payment_method_data,
           }});
       },
-    handleCardPayment: async (clientSecret, element, data) => {
-      console.log('localstripe: Stripe().handleCardPayment()');
+    confirmCardPayment: async (clientSecret, data) => {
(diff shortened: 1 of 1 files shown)


localstripe/localstripe-v3.js
```diff
@@ -212,101 +309,14 @@ Stripe = (apiKey) => {
...
-      } catch (err) {
-        if (typeof err === 'object' && err.error) {
-          return err;
-        } else {
-          return {error: err};
-        }
-      }
+    confirmCardSetup: confirmCardSetup,
+    handleCardPayment: async (clientSecret, element, data) => {  // deprecated
+      return confirmCardPayment(clientSecret);
```
adrienverge: Isn't it needed to pass some `data` here?

H--o-l: `data` is part of the function prototype, but it's not use.
It was already like that before, and it's the case for `handleCardSetup` where data is ignored:
```js
    handleCardSetup: async (clientSecret, element, data) => {  // deprecated
      return confirmCardSetup(clientSecret, {payment_method: {card: element}});
    },
```
Do you think it's a bad idea?

H--o-l:
> To minimize the diff and for clearer code, can you use:
> 
> ```ts
> Stripe = (apiKey) => {
>   return {
>     elements: // ...
>     createToken: // ...
>     // ...
>     confirmCardPayment: async function (clientSecret, data) {
>       console.log('localstripe: Stripe().confirmCardPayment()');
>       // ...
>     },
>     handleCardPayment: async function (clientSecret, element, data) {  // deprecated
>       return this.confirmCardPayment(clientSecret);
>     },
>     // ...
> ```
> 
> ?

It's better indeed, done, thanks!

H--o-l:
Hey @adrienverge.
I have push a new version following your proposal here https://github.com/adrienverge/localstripe/pull/111#pullrequestreview-333167398 but in fact I can't make it work, I have the following error: `this.confirmCardSetup is not a function`.
With the little knowledge I have in Javascript, I don't know how to make it works like you propose.
I think I should revert to my first version.
What do you think?

localstripe/localstripe-v3.js
adrienverge: OK! Let's ignore it.

adrienverge:
It doesn't work because you didn't take *exactly* my code: if you don't declare a function with `function`, you create an *arrow function*, for which the keyword `this` doesn't have the same value :wink: Can you retry with my proposal?

H--o-l:
Crap, you are right, thanks!
I had tried both `function` and `arrow function`, but on the new functions, not the old ones...
So here is the diff. It's still not your solution because I used `function` only for the old functions, it seems more logic to me, in regards to the rest of the file.
```patch
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -277,7 +277,8 @@ Stripe = (apiKey) => {
         }
       }
     },
-    handleCardSetup: async (clientSecret, element, data) => {  // deprecated
+    handleCardSetup:  // deprecated
+        async function (clientSecret, element, data) {
       return this.confirmCardSetup(clientSecret,
                                    {payment_method: {card: element}});
     },
@@ -311,7 +312,8 @@ Stripe = (apiKey) => {
         }
       }
     },
-    handleCardPayment: async (clientSecret, element, data) => {  // deprecated
+    handleCardPayment:  // deprecated
+        async function (clientSecret, element, data) {
       return this.confirmCardPayment(clientSecret);
     },

File: localstripe/localstripe-v3.js [hunk 2]
```diff
@@ -226,13 +226,13 @@ Stripe = (apiKey) => {
             payment_method_data: {
               type: 'card',
               card: {
-                number: element.value.number,
-                exp_month: element.value.exp_month,
-                exp_year: element.value.exp_year,
-                cvc: element.value.cvc,
+                number: data.payment_method.card.value.number,
```
adrienverge: To keep code consistent, and to simplify, could we do the same as for SEPA?

https://github.com/adrienverge/localstripe/blob/c11d14df298cfb271473ad62534636000a29ba5a/localstripe/localstripe-v3.js#L324

localstripe/localstripe-v3.js
```diff
@@ -226,13 +226,13 @@ Stripe = (apiKey) => {
...
+                exp_month: data.payment_method.card.value.exp_month,
+                exp_year: data.payment_method.card.value.exp_year,
+                cvc: data.payment_method.card.value.cvc,
               },
               billing_details: {
-                address: {postal_code: element.value.address_zip},
+                address: {postal_code: data.payment_method.card.value.address_zip},
```
adrienverge: From my understanding of the new `confirmCardSetup()` API, the `data` argument doesn't have a `.payment_method.card.value.address_zip`. Instead, it should be called with `.payment_method` being an "object containing data to create a PaymentMethod with", like in https://stripe.com/docs/api/payment_methods/create.

It would mean:
- This code would crash if called from code that supports the new Stripe arguments?
- This seems to go in favor of my comment just above!
- Modifications are probably needed in the *card mounting* code, above in the file.
- Modifications are probably need in legacy `handleCardSetup()`, so that old code keeps worknig.

H--o-l: Hey @adrienverge, sorry but it have taken me forever to only start to understand.
What you mean is that you think they have change the object returned by https://stripe.com/docs/js/elements_object/create_element?type=card (and properties added in it by `mount`) in the same time they depreciate `handleCardSetup()`?

adrienverge: (I know, it also took me a long time (reading the docs) to detect this potential problem!)

No, I'm saying that when `Stripe().confirmCardSetup()` is called from a valid JS app, it is now called with a `data` argument (no more `Element`/card object passed as direct argument), and this `data` argument doesn't seem to contain `data.payment_method.card.value.address_zip` (correct me if I'm wrong, but this is what I understand from the docs).

So, the function as it is written now, won't work with:

```js
// example from the docs
Stripe().confirmCardSetup(clientSecret, {
  payment_method: {
    card: cardElement,
    billing_details: {
      address: {
        postal_code: '12345',
      },
    },
  },
});
```

neither with:

```js
// example from the docs
Stripe().confirmCardSetup(clientSecret, {
  payment_method: {
    type: "card",
    card: {
      "number": "4242424242424242",
      "exp_month": 12,
      "exp_year": 2020,
      "cvc": "314",
    },
    billing_details: {
      address: {
        postal_code: '12345',
      },
    },
  },
});
```

because it will:
- ignore `billing_details.address.postal_code`
- crash on `data.payment_method.card.value.address_zip` (actually it won't crash today, because of the old *card mounting code*, cf. my previous comment).

Do you confirm my reading of the documentation?

H--o-l: Hey Adrien.
I have look more into the `Element` object, in my console using the real Stripe API and inside [@type/stripe-v3](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/stripe-v3/index.d.ts#L671-L685).
For me all the variables we use here to simulate the API (`card.value.number`, `card.value.exp_year`, etc) don't exists at all in real situation.
Thus, it can't:
- lead to a crash
- make old code unworking.
No?
I can simplify the object so it better glue with the final http fetch to the http API, but in another commit/PR, no?

localstripe/localstripe-v3.js
H--o-l: Hey Adrien!
So It's not possible to do that because when we follow [this way of using the API](https://stripe.com/docs/payments/save-and-reuse#add-elements-to-your-page) we have a 
```js
data.payment_method.card = {
   listeners: {change: [null]},
   values: {number: 12, exp_month:12, ...} `
```
And it's not directcly compatible with the API.
Will you be OK if I had a condition to check if the user used an object directly, or a Stripe Element?

H--o-l:
Hey Adrien!
I added a new commit before the other following your review.
I think it handles examples you were mentioning, and that it's as close as possible to what's Stripe is probably doing.
However it's really tricky, I tested a lot of variations before having something that works.
I let you see!

localstripe/localstripe-v3.js
```diff
@@ -277,8 +275,16 @@ Stripe = (apiKey) => {
         }
       }
     },
-    handleCardPayment: async (clientSecret, element, data) => {
-      console.log('localstripe: Stripe().handleCardPayment()');
+    handleCardSetup:  // deprecated
+      async function (clientSecret, element, data) {
+        return this.confirmCardSetup(clientSecret, {
+          payment_method: {
+            card: element.blackbox,
```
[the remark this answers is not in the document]
H--o-l: That's possible indeed, I hesitate, It wasn't changing much.
If you think it's clearer I'm happily changing it!

localstripe/localstripe-v3.js
```diff
@@ -98,29 +98,32 @@ class Element {
...
         exp_month: inputs.exp_month.value,
         exp_year: '20' + inputs.exp_year.value,
         cvc: inputs.cvc.value,
-        address_zip: inputs.address_zip.value,
-      };
+      }
+
+      this.value = {
+        postal_code: inputs.postal_code.value,
+      }
```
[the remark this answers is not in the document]
H--o-l: Hey!
I use a `this.value.postal_code` structure to be closer to
- https://stripe.com/docs/js/elements_object/create_element?type=card#elements_create-options-value
and
- https://stripe.com/docs/js/element/events/on_change?type=cardElement#element_on_change-handler-value

Also, I think the `this.value` could be used to store things like `name` in the futur if the user want's it.

What do you think?

localstripe/localstripe-v3.js
```diff
@@ -225,15 +228,8 @@ Stripe = (apiKey) => {
...
-              card: {
-                number: element.value.number,
-                exp_month: element.value.exp_month,
-                exp_year: element.value.exp_year,
-                cvc: element.value.cvc,
-              },
-              billing_details: {
-                address: {postal_code: element.value.address_zip},
-              },
+              card: element.blackbox ? element.blackbox: element,
```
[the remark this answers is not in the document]
H--o-l: Thanks!

H--o-l:
> Yes, and it would be better, good idea! Something like card instanceof Element?

:+1: 

> The current implementation still doesn't take the postal code from the element, to put it inside billing_details.

Hey!
I think my first version allowed it, from our Element.
This one however did not indeed, and I said it inside the commit message:
```
User will have to give postal code like that now:

Stripe().handleCardSetup(clientSecret, {
  element,
  payment_method_data: {
    billing_details: {
      address: {
        postal_code: element.value.postal_code,
      },
    },
  },
});
```


> Would the following proposal work?
> 
> ```js
>     confirmCardSetup: async (clientSecret, data) => {
>       // ...
>         if (data.payment_method.card instanceof Element) {
>           const element = data.payment_method.card;
>           data.payment_method.card = element.card;
>           data.payment_method.billing_details =
> 	    data.payment_method.billing_details || {};
>           data.payment_method.billing_details.address =
>             data.payment_method.billing_details.address || {};
>           data.payment_method.billing_details.address.postal_code =
>             data.payment_method.billing_details.address.postal_code ||
> 	    element.postal_code;
>         }
> ```

I will indeed! And it will fix the reading of the postal code from Element.
Thanks for the proposition, I have integrated in first and second commits!

I adapt the first commit message text, and redo manual test with a card, for the old and the new API.

localstripe/localstripe-v3.js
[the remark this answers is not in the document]
adrienverge: I do think it's clearer, but also simpler code, and more forward-compatible!

localstripe/localstripe-v3.js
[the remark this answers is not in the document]
adrienverge: > Edit: I like the following structure, I will use it if it's OK for you!

Perfect! I like it too :+1:

localstripe/localstripe-v3.js
```diff
@@ -215,22 +215,21 @@ Stripe = (apiKey) => {
...
+        if (data.payment_method.card instanceof Element) {
+          const element = data.payment_method.card;
+          data.payment_method.card = element.value.card;
+          data.payment_method.billing_details =
+            data.payment_method.billing_details || {};
+          data.payment_method.billing_details.address =
+            data.payment_method.billing_details.address || {};
+          data.payment_method.billing_details.address.postal_code =
+            data.payment_method.billing_details.address.postal_code ||
+          element.value.postal_code;
```
adrienverge: Indentation problem!

H--o-l: :eagle: :eyes: thanks, fixed!
```diff
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -229,7 +229,7 @@ Stripe = (apiKey) => {
             data.payment_method.billing_details.address || {};
           data.payment_method.billing_details.address.postal_code =
             data.payment_method.billing_details.address.postal_code ||
-          element.value.postal_code;
+            element.value.postal_code;
         }
         let response = await fetch(url, {
           method: 'POST',

PR #114, opened 2020-01-03
Remove from property in plan tiers to match API definition

marotpam:
Long overdue fix for issue reported in https://github.com/adrienverge/localstripe/issues/44. As per the official API [docs](https://stripe.com/docs/api/plans/object#plan_object-tiers-up_to), plan tiers only have the `up_to` property, not `from`.
With this change, when calculating plan amounts, the from will be initially set to `0` for the first tier, and the `up_to` value of the previous tier + 1 for the rest of tiers


adrienverge:
@rasmuscnielsen I understand from your emoji to https://github.com/adrienverge/localstripe/pull/114#issue-359032903 that it's working for you too. Hence I'll merge. @marotpam thanks for your contribution!

PR #119, opened 2020-02-10
feat: Support invoice line items (il_) from Stripe API 2019-12-03

bravier:
Stripe API is returning `InvoiceLineItem`s in its latest version
instead of `InvoiceItem`s as it may have in the past.

This new object acts as a representation of either an
`InvoiceItem` or a `SubscriptionItem` for each line of an `Invoice`.

More info on [Stripe's documentation](https://stripe.com/docs/api/invoices/line_item).

Let's support them in localstripe!

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -912,7 +912,7 @@ class Invoice(StripeObject):
         self.lines = List('/v1/invoices/' + self.id + '/lines')
         for item in items:
             item.invoice = self.id
-            self.lines._list.append(item)
+            self.lines._list.append(InvoiceLineItem(item))
 
         pending_items = [ii for ii in InvoiceItem._api_list_all(
             None, customer=self.customer, limit=99)._list
@@ -920,7 +920,7 @@ class Invoice(StripeObject):
         for ii in pending_items:
             if not simulation:
                 ii.invoice = self.id
-            self.lines._list.append(ii)
+            self.lines._list.append(InvoiceLineItem(ii))
 
         if len(self.lines._list):
             self.currency = self.lines._list[0].currency
@@ -935,7 +935,7 @@ class Invoice(StripeObject):
 
     @property
     def subtotal(self):
-        return sum([ii.amount for ii in self.lines._list])
+        return sum([il.amount for il in self.lines._list])
 
     @property
     def tax(self):
@@ -947,12 +947,12 @@ class Invoice(StripeObject):
     @property
     def total_tax_amounts(self):
         concat = []
-        for ii in self.lines._list:
+        for il in self.lines._list:
             tax_amounts = []
-            if ii.tax_rates:
-                tax_amounts = ii.tax_amounts
+            if il.tax_rates:
+                tax_amounts = il.tax_amounts
             elif self.default_tax_rates:
-                tax_amounts = [tr._tax_amount(ii.amount)
+                tax_amounts = [tr._tax_amount(il.amount)
                                for tr in self.default_tax_rates]
             concat.extend(tax_amounts)
         # TODO: reduce `concat` by unique `tax_rate` ID
@@ -1189,9 +1189,9 @@ class Invoice(StripeObject):
                           simulation=True)
 
             if subscription_proration_date is not None:
-                for ii in invoice.lines._list:
-                    ii.period['start'] = subscription_proration_date
-                    ii.period['end'] = subscription_proration_date
+                for il in invoice.lines._list:
+                    il.period['start'] = subscription_proration_date
+                    il.period['end'] = subscription_proration_date
 
             return invoice
 
@@ -1304,11 +1304,24 @@ class Invoice(StripeObject):
 
         return obj
 
+    @classmethod
+    def _api_list_lines(cls, id, limit=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        obj = cls._api_retrieve(id)
+
+        lines = List('/v1/invoices/' + id + '/lines', limit=limit)
+        lines._list = obj.lines._list
+
+        return lines
+
 
 extra_apis.extend((
     ('GET', '/v1/invoices/upcoming', Invoice._api_upcoming_invoice),
     ('POST', '/v1/invoices/{id}/pay', Invoice._api_pay_invoice),
-    ('POST', '/v1/invoices/{id}/void', Invoice._api_void_invoice)))
+    ('POST', '/v1/invoices/{id}/void', Invoice._api_void_invoice),
+    ('GET', '/v1/invoices/{id}/lines', Invoice._api_list_lines)))
 
 
 class InvoiceItem(StripeObject):
@@ -1368,6 +1381,7 @@ class InvoiceItem(StripeObject):
         self.invoice = invoice
         self.subscription = subscription
         self.plan = plan
+        self.quantity = 1
         self.amount = amount
         self.currency = currency
         self.customer = customer
@@ -1377,11 +1391,6 @@ class InvoiceItem(StripeObject):
         self.description = description
         self.tax_rates = tax_rates
         self.metadata = metadata or {}
-
-    @property
-    def tax_amounts(self):
-        if self.tax_rates is not None:
-            return [tr._tax_amount(self.amount) for tr in self.tax_rates]
 
     @classmethod
     def _api_list_all(cls, url, customer=None, limit=None):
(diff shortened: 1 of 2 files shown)


localstripe/resources.py
```diff
@@ -1400,6 +1395,70 @@ def _api_list_all(cls, url, customer=None, limit=None):
...
+        self.metadata = item.metadata
+
+        self.amount = item.amount \
+            if self.type == 'invoiceitem' else item._calculate_amount()
+        self.currency = \
+            item.currency if self.type == 'invoiceitem' else item.plan.currency
+        self.description = \
+            item.description if self.type == 'invoiceitem' else item.plan.name
+        self.proration = \
+            item.proration if self.type == 'invoiceitem' else False
```
bravier: `proration` can only exist when the `InvoiceLineItem` references an `InvoiceItem`.
Using `False` as the default value here for `SubscriptionItem`s.

localstripe/resources.py
```diff
@@ -1400,6 +1395,70 @@ def _api_list_all(cls, url, customer=None, limit=None):
...
+
+        self.quantity = item.quantity if self.type == 'subscription' else 1
```
bravier: In `localstripe` the quantity can only be set on `SubscriptionItem`s so I use `1` as the default value for `InvoiceItem`s.

localstripe/resources.py
```diff
@@ -1378,11 +1378,6 @@ def __init__(self, invoice=None, subscription=None, plan=None, amount=None,
         self.tax_rates = tax_rates
         self.metadata = metadata or {}
 
-    @property
-    def tax_amounts(self):
-        if self.tax_rates is not None:
-            return [tr._tax_amount(self.amount) for tr in self.tax_rates]
-
```
adrienverge: :+1:

localstripe/resources.py
```diff
@@ -2304,16 +2363,7 @@ def _create_invoice(self):
...
-                InvoiceItem(subscription=self.id, plan=si.plan.id,
-                            amount=si._calculate_amount(),
-                            currency=si.plan.currency,
-                            description=si.plan.name,
-                            tax_rates=[tr.id
-                                       for tr in (si.tax_rates or [])],
-                            customer=self.customer,
-                            period_start=self.current_period_start,
-                            period_end=self.current_period_end))
+            pending_items.append(si)
```
adrienverge: The code looks good, but can it be a problem to reuse the same objects here (instead of creating new objects, like the previous code did)? @H--o-l I think you reworked this code not so long ago (commit 5fca63eb), do you have an opinion?

localstripe/resources.py
adrienverge: Perfect! It seems in agreement with what Stripe API doc examples show.

localstripe/resources.py
```diff
@@ -1400,6 +1395,70 @@ def _api_list_all(cls, url, customer=None, limit=None):
...
+
+        self.type = \
+            'invoiceitem' if isinstance(item, InvoiceItem) else 'subscription'
+
+        if self.type == 'subscription':
+            self.subscription_item = item.id
+            self.subscription = item._subscription
+        elif self.type == 'invoiceitem':
+            self.invoice_item = item.id
+            self.subscription = None
```
adrienverge: Maybe I suggest the following? It seems more readable to me:
```python
        if self.type == 'subscription':
            self.subscription_item = item.id
            self.subscription = item._subscription
            self.plan = item.plan
            self.quantity = item.quantity
            self.amount = item._calculate_amount()
            self.currency = item.plan.currency
            ...
            subscription_obj = Subscription._api_retrieve(item._subscription)
            self.period = dict(start=subscription_obj.current_period_start,
                               end=subscription_obj.current_period_end)
        elif self.type == 'invoiceitem':
            self.invoice_item = item.id
            self.subscription = None
            self.plan = None
            self.quantity = 1
            self.currency = item.currency
            ...
            self.amount = item.amount
            self.period = item.period
```

localstripe/resources.py
bravier: Please note that `InvoiceItem` has a `quantity` field (see the [doc](https://stripe.com/docs/api/invoiceitems/object#invoiceitem_object-quantity)) but localstripe do not implement it. Maybe a comment in the code can help better understand this default value?

adrienverge: Ah, thanks, I hadn't seen that!

Maybe a semi-quick-win would be to add this in `InvoiceItem`:
```diff
@@ -1368,6 +1368,7 @@ class InvoiceItem(StripeObject):
         self.invoice = invoice
         self.subscription = subscription
         self.plan = plan
+        self.quantity = 1
         self.amount = amount
         self.currency = currency
         self.customer = customer
```
and just use this in `InvoiceLineItem`:
```python
self.quantity = item.quantity
```
?

localstripe/resources.py
bravier: Indeed it is. Done!

localstripe/resources.py
bravier: Good idea. It would be clearer on why this number is 1 by default. Code updated.

localstripe/resources.py
H--o-l: Hey!

For me it looks OK, because a new object is not created here, but will be at this line https://github.com/adrienverge/localstripe/pull/119/files#diff-62ee820ebbbef6539c97fecb3a10c374R915

Object that are created and should not be removed without breaking something are :
https://github.com/adrienverge/localstripe/blob/c6821928953c50058327f04ef11b9e70bc8e609b/localstripe/resources.py#L2471-L2476
These end-ups being the one we get with:
```py
        pending_items = [ii for ii in InvoiceItem._api_list_all(
            None, customer=customer, limit=99)._list
            if ii.invoice is None]
```
TL;DR: OK for me! @baptsite as said on Slack, unit test could be cool, and the following API https://stripe.com/docs/api/invoices/invoice_lines too otherwise OK for me.

adrienverge: @H--o-l thanks :+1:

bravier:
Hi Adrien,
Small update to let you know I've added support for retrieving an invoice's line item on a specific route (https://stripe.com/docs/api/invoices/invoice_lines) and a small test to verify this route exists on an invoice.
I noticed parameters `ending_before` and `starting_after` are not implemented for other routes which might support them so I didn't implemented them here neither. Hope it's ok like that.

localstripe/resources.py
```diff
@@ -1304,11 +1304,24 @@ def _api_void_invoice(cls, id):
...
 
+    @classmethod
+    def _api_list_lines(cls, id, limit=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        obj = cls._api_retrieve(id)
+
+        lines = List('/v1/invoices/' + id + '/lines', limit=limit)
+        lines._list = obj.lines._list
```
adrienverge: I'd be something simpler, like in
https://github.com/adrienverge/localstripe/blob/c6821928953c50058327f04ef11b9e70bc8e609b/localstripe/resources.py#L749-L753
or
https://github.com/adrienverge/localstripe/blob/c6821928953c50058327f04ef11b9e70bc8e609b/localstripe/resources.py#L741-L746

Is it possible, or is it a hidden reason for the need to create a new `List`?

test.sh
```diff
@@ -399,11 +399,12 @@ curl -sSf -u $SK: $HOST/v1/subscriptions \
...
-           -d customer=$cus \
-           -d items[0][plan]=basique-mensuel \
-           -d items[0][tax_rates][0]=$txr1 \
-      | grep -oE 'sub_\w+' | head -n 1)
+res=$(curl -sSf -u $SK: $HOST/v1/subscriptions \
+                    -d customer=$cus \
+                    -d items[0][plan]=basique-mensuel \
+                    -d items[0][tax_rates][0]=$txr1)
+sub=$(echo "$res" | grep -oE 'sub_\w+' | head -n 1)
+inv=$(echo "$res" | grep -oE 'in_\w+' | head -n 1)
```
adrienverge: This is nitpicking, but can you use `in`, to stay within Stripe prefixes?

localstripe/resources.py
bravier: It was mainly to make use of the `List` class validation code for the `limit` parameter instead of copying it.

adrienverge: Ah, OK! It's a good reason.

test.sh
bravier: Code updated:
```diff
--- a/test.sh
+++ b/test.sh
@@ -404,7 +404,7 @@ res=$(curl -sSf -u $SK: $HOST/v1/subscriptions \
                     -d items[0][plan]=basique-mensuel \
                     -d items[0][tax_rates][0]=$txr1)
 sub=$(echo "$res" | grep -oE 'sub_\w+' | head -n 1)
-inv=$(echo "$res" | grep -oE 'in_\w+' | head -n 1)
+in=$(echo "$res" | grep -oE 'in_\w+' | head -n 1)
 
 curl -sSf -u $SK: $HOST/v1/invoices?customer=$cus
 
@@ -414,7 +414,7 @@ curl -sSf -u $SK: $HOST/v1/invoices/upcoming?customer=$cus\&subscription_items%5
 
 curl -sSf -u $SK: $HOST/v1/invoices/upcoming?customer=$cus\&subscription=$sub\&subscription_items%5B0%5D%5Bid%5D=si_RBrVStcKDimMnp\&subscription_items%5B0%5D%5Bplan%5D=basique-annuel\&subscription_proration_date=1504182686\&subscription_tax_percent=20
 
-curl -sSf -u $SK: $HOST/v1/invoices/$inv/lines
+curl -sSf -u $SK: $HOST/v1/invoices/$in/lines
 
 cus=$(curl -sSf -u $SK: $HOST/v1/customers \
            -d description='This customer will have a subscription with volume tiered pricing' \
```

adrienverge:
Perfect!

PR #133, opened 2020-04-10
Added off_session and confirm to PaymentIntent creation

### LINKED ISSUE #130: Unable to create PaymentIntent
Hello,
I'm using off_session charges to process subscriptions on my site. Problem is, when I use code such as:
```php
$payment = \Stripe\PaymentIntent::create([
    (...)
    'off_session' => true,
    'confirm' => true,
]);
```

I get an error regarding off_session and confirm parameters to be unexcepted.
Exact error: `Stripe\Exception\InvalidRequestException : Unexpected off_session, confirm`.
At first, I thought that it is lib-related problem, but that's not the case (switching to regular stripe API proves that code is correct as the intent is being created without any problem).

I would highly appreciate any help regarding that issue.

Best regards.

ImJustToNy:
Closes #130

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1511,11 +1511,14 @@ class PaymentIntent(StripeObject):
     _id_prefix = 'pi_'
 
     def __init__(self, amount=None, currency=None, customer=None,
-                 payment_method=None, metadata=None, **kwargs):
+                 payment_method=None, metadata=None, off_session=None,
+                 confirm=None, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
         amount = try_convert_to_int(amount)
+        off_session = try_convert_to_bool(off_session)
+        confirm = try_convert_to_bool(confirm)
         try:
             # Invoices with amount == 0 don't create PaymentIntents:
             assert type(amount) is int and amount > 0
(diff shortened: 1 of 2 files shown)


ImJustToNy:
Actually, it seems to me that `off_session` and `confirm` can be used separately, so I'm not so sure if 

> Check that off_session is true only if confirm is true too.

is good thing to do.

localstripe/resources.py
```diff
@@ -1527,6 +1530,8 @@ def __init__(self, amount=None, currency=None, customer=None,
                 assert (payment_method.startswith('pm_') or
                         payment_method.startswith('src_') or
                         payment_method.startswith('card_'))
+            assert type(off_session) is bool
+            assert type(confirm) is bool
```
adrienverge, line 1534: You need to
```suggestion
            if confirm is not None:
                assert type(confirm) is bool
            if off_session is not None:
                assert type(off_session) is bool
```
Otherwise current requests (without passing `confirm`) won't work anymore, and many users will complain.

ImJustToNy:
> >  so I'm not so sure if [...] is good thing to do.
> 
> In [Stripe API documentation](https://stripe.com/docs/api/payment_intents/create), I read:
> 
> > `off_session` [optional] [ONLY WHEN CONFIRM=TRUE]
> Set to `true` to indicate that the customer is not in your checkout flow during this payment attempt, and therefore is unable to authenticate. This parameter is intended for scenarios where you collect card details and charge them later. This parameter can only be used with `confirm=true`.
> 
> Am I misunderstanding it?

Turns out that I misunderstood documentation 😅

ImJustToNy:
I've added some simple test for that assertion, I hope it's correct :)

PR #135, opened 2020-04-23
Confirm parameter on payment intent should trigger payment

ImJustToNy:


localstripe/resources.py
```diff
@@ -1561,6 +1561,9 @@ def __init__(self, amount=None, currency=None, customer=None,
         self._canceled = False
         self._authentication_failed = False
 
+        if confirm:
+            self._trigger_payment()
```
adrienverge, line 1565: Will this work if the payment method requires authentication? Maybe it's better to reuse function `PaymentIntent._api_confirm(self)`, that does all needed checks?

> When confirm=true is used during creation, it is equivalent to creating and confirming the PaymentIntent in the same call.

--- from https://stripe.com/docs/api/payment_intents/create

PR #136, opened 2020-04-25
PaymentIntent: Refactor passing of confirm and off_session

adrienverge:
`confirm` and `off_session` are not part of the object itself. They are
read only on *creation*, and used only once to confirm the intent:

From https://stripe.com/docs/api/payment_intents/create:
> When `confirm=true` is used during creation, it is equivalent to
> creating and confirming the PaymentIntent in the same call.
> You may use any parameters available in the confirm API when
> `confirm=true` is supplied.

---

@ImJustToNy this follows #133 and #135. Can you review it or test it? Thanks!

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1512,14 +1512,11 @@ class PaymentIntent(StripeObject):
     _id_prefix = 'pi_'
 
     def __init__(self, amount=None, currency=None, customer=None,
-                 payment_method=None, metadata=None, off_session=None,
-                 confirm=None, **kwargs):
+                 payment_method=None, metadata=None, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
         amount = try_convert_to_int(amount)
-        off_session = try_convert_to_bool(off_session)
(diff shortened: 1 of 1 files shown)


localstripe/resources.py
```diff
@@ -1625,6 +1614,26 @@ def last_payment_error(self):
...
+            if off_session is not None:
+                assert type(off_session) is bool
+                assert confirm is True
+        except AssertionError:
+            raise UserError(400, 'Bad request')
+
+        obj = super()._api_create(**data)
+
+        if confirm:
+            cls._api_confirm(obj)
```
ImJustToNy, line 1633: We should `_api_confirm` with ID instead of entire object.

```python
if confirm:
    cls._api_confirm(obj.id)
```

adrienverge: Ah, of course! Thanks.

Updated:
```diff
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1630,7 +1630,7 @@ class PaymentIntent(StripeObject):
         obj = super()._api_create(**data)
 
         if confirm:
-            cls._api_confirm(obj)
+            cls._api_confirm(obj.id)
 
         return obj
 
```

PR #138, opened 2020-04-25
Subscription: Correctly set latest_invoice on manual payment

adrienverge:
This fixes the following bug: if a subscription was upgraded to a plan
with the same billing cycle (e.g. yearly → yearly), and the user
generated the invoice manually, then the `Subscription.latest_invoice`
wasn't updated to the new value.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -841,7 +841,7 @@ class Invoice(StripeObject):
 
     def __init__(self, customer=None, subscription=None, metadata=None,
                  items=[], date=None, description=None,
-                 simulation=False,
+                 simulation=False, upcoming=False,
                  tax_percent=None,  # deprecated
                  default_tax_rates=None,
                  **kwargs):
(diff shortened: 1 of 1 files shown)


localstripe/resources.py
```diff
@@ -935,7 +935,10 @@ def __init__(self, customer=None, subscription=None, metadata=None,
         self._draft = True
         self._voided = False
 
-        if not simulation:
+        if not simulation and not upcoming:
+            if subscription is not None:
+                subscription_obj.latest_invoice = self.id
+
             schedule_webhook(Event('invoice.created', self))
```
H--o-l, line 942: A small question: the condition to execute this line seems to have changed (condition `not upcoming` was added).
Thus it changes when this webhook is send?

adrienverge: Yes, totally right!

Before, an `invoice.created` event was sent the user looked at the next upcoming invoice (`GET /v1/invoices/upcoming?customer=cus_n2FydPZCLKMDXk`).
Now, it's not the case anymore (and I think it makes more sense, but feel free to comment if I'm wrong).

H--o-l: Indeed it make sense, thanks for the answer!

PR #142, opened 2020-07-01
convert: add double convert to int on error

sams96:
This allows converting values like `"50.0"` which is what stripe-go will post for the `percent_off` value in POST coupon.

Sorry I'm not much of a python guy, so I'm not sure if the double try is very idiomatic. I'd be happy to change it if you have a better suggestion.

Thanks

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -518,7 +518,7 @@ class Coupon(StripeObject):
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
         amount_off = try_convert_to_int(amount_off)
-        percent_off = try_convert_to_int(percent_off)
+        percent_off = try_convert_to_float(percent_off)
         duration_in_months = try_convert_to_int(duration_in_months)
         try:
             assert type(id) is str and id


adrienverge:
Hi Sam, thanks for your contribution!

Since the [docs are explicit](https://stripe.com/docs/api/coupons/create#create_coupon-percent_off) about floats being accepted on `POST /v1/coupons`, I'd be more in favor of using `try_convert_to_float` at https://github.com/adrienverge/localstripe/blob/597e4dde7a6cef31eded345c9040d2b62e927e6f/localstripe/resources.py#L521
This way, `try_convert_to_int` would still reject floats in other functions.

What do you think?

sams96:
Yes that makes a lot of sense, I have replaced my commit with one that makes this change instead.

sams96:
Also I have made quite a few other changes in my fork's master (added name field to coupons, added discounts object, added trailing status to subscriptions), and I will be making some more. I figured it was easier for me to just use my fork for now, but I'd be happy to make PRs for them (or one PR with everything) if you want.

adrienverge:
If your other changes reflect real Stripe behavior, and don't introduce bugs, they are welcome here :)
(one pull request per change is easier to review)

PR #144, opened 2020-07-07
fix(resources): Made charges list API parameters optionnal

H--o-l:
PR based on previous PR https://github.com/adrienverge/localstripe/pull/143

--- 
### fix(resources): Made charges list API parameters optionnal

In Stripe [documentation of the Charges list API], all parameters are optional.
Since commit 54831f9 _Charge: Add `customer` and `created` to the list API_,
localstripe support parameters `customer` and `created` but both were made
mandatory.
Let's fix them.

[documentation of the Charges list API]: https://stripe.com/docs/api/charges/list


localstripe/resources.py
```diff
@@ -478,27 +478,31 @@ def refunded(self):
...
-            assert _type(customer) is str and customer.startswith('cus_')
-            assert type(created) in (dict, str)
-            if type(created) is dict:
-                assert len(created.keys()) == 1 and \
-                    list(created.keys())[0] in ('gt', 'gte', 'lt', 'lte')
-                date = try_convert_to_int(list(created.values())[0])
-            elif type(created) is str:
-                date = try_convert_to_int(created)
-            assert type(date) is int and date > 1500000000
+            if customer:
```
adrienverge, line 481: To be consistent with the rest of the code, can you use `is not None` (just here, not needed after `except AssertionError:`)?

Otherwise the API would accept values like `""` and `0`.

H--o-l: Ok indeed good idea!
Here is the diff:
```diff
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -478,9 +478,9 @@ class Charge(StripeObject):
     @classmethod
     def _api_list_all(cls, url, customer=None, created=None, limit=10):
         try:
-            if customer:
+            if customer is not None:
                 assert _type(customer) is str and customer.startswith('cus_')
-            if created:
+            if created is not None:
                 assert type(created) in (dict, str)
                 if type(created) is dict:
                     assert len(created.keys()) == 1 and \

File: localstripe/resources.py [hunk 2]
```diff
@@ -478,27 +478,31 @@ def refunded(self):
...
-                assert len(created.keys()) == 1 and \
-                    list(created.keys())[0] in ('gt', 'gte', 'lt', 'lte')
-                date = try_convert_to_int(list(created.values())[0])
-            elif type(created) is str:
-                date = try_convert_to_int(created)
-            assert type(date) is int and date > 1500000000
+            if customer is not None:
+                assert _type(customer) is str and customer.startswith('cus_')
+            if created is not None:
+                assert type(created) in (dict, str)
```
adrienverge, line 484: Not directly related to this PR, but there is a mix of `_type()` and `type()` here. Can you fix it in the same time?

H--o-l: Crap indeed thanks, here is the fix:
```diff
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -481,14 +481,14 @@ class Charge(StripeObject):
             if customer is not None:
                 assert _type(customer) is str and customer.startswith('cus_')
             if created is not None:
-                assert type(created) in (dict, str)
-                if type(created) is dict:
+                assert _type(created) in (dict, str)
+                if _type(created) is dict:
                     assert len(created.keys()) == 1 and \
                         list(created.keys())[0] in ('gt', 'gte', 'lt', 'lte')
                     date = try_convert_to_int(list(created.values())[0])
-                elif type(created) is str:
+                elif _type(created) is str:
                     date = try_convert_to_int(created)
-                assert type(date) is int and date > 1500000000
+                assert _type(date) is int and date > 1500000000
         except AssertionError:
             raise UserError(400, 'Bad request')
 
@@ -496,7 +496,7 @@ class Charge(StripeObject):
             Customer._api_retrieve(customer)  # to return 404 if not existant
 
         if created:
-            if type(created) is str or not created.get('gt'):
+            if _type(created) is str or not created.get('gt'):
                 raise UserError(500, 'Not implemented')
 
         li = super(Charge, cls)._api_list_all(url, limit=limit)

adrienverge: Hum, can you do the opposite? There's no need to use a custom `_type()` here, I believe.

H--o-l: Oh, I wasn't carreful the custom `_type()` was used only when necessary.
I switch it back and fix the first one. Thanks!

H--o-l:
Thanks for the review and the help!

Issue #88, opened 2019-08-23
Can't create a Charge


matbarker:
Trying to create a charge using the docker image fails with a generic error message:

```
#!/bin/bash
set -eux

HOST=http://localhost:8420
SK=sk_test_12345
	 
curl -sSf -u $SK: $HOST/v1/charges \
  -d amount=2000 \
  -d currency=usd \
  -d source=tok_visa
```
Response:

```HTTP/1.1 400 Bad Request
Content-Type: application/json; charset=utf-8
Content-Length: 87
Date: Thu, 22 Aug 2019 10:03:27 GMT
Server: Python/3.7 aiohttp/3.5.4

{
  "error": {
    "message": "Bad request",
    "type": "invalid_request_error"
  }
}
```

How do I get a more specific error message, or figure out what I'm doing wrong?

adrienverge:
Hi @matbarker,

To know what's going wrong, the best is probably to look at [the `Charge` class definition](https://github.com/adrienverge/localstripe/blob/8e84d1835a37cc2ccac9f867e70a37b2a5bb76f0/localstripe/resources.py#L317-L318). I think what's failing here is:
https://github.com/adrienverge/localstripe/blob/8e84d1835a37cc2ccac9f867e70a37b2a5bb76f0/localstripe/resources.py#L332-L333

- Currently, localstripe only accept *sources*, not *card tokens*. The "supported" use case is to attach the token to a customer first, in order to have a `card_` object:
  ```shell
  curl -sSf -u $SK: localhost:8420/v1/customers/cus_zLHwf2fPMEmIxF/sources
       -d source=tok_1oBvNLpoJRZ3xP
  ```
  Of course, contributions to support creating charges directly from tokens are welcome.

- localstripe acts like *real* Stripe servers, which means example tokens from the docs (like `tok_visa`) don't work. You need to create a token, like in real-life conditions:
  ```shell
  curl -sSf localhost:8420/v1/tokens \
       -d key=pk_test_sldkjflaksdfj \
       -d card[number]=4242424242424242 \
       -d card[exp_month]=12 \
       -d card[exp_year]=2019 \
       -d card[cvc]=123
  ```

PR #148, opened 2020-08-07
This is one step in fixing issue #88

meatherly:
It implements the `GET '/v1/customers/{id}/sources'` endpoint for a customer as described here: https://stripe.com/docs/api/cards/list

Also I've never written python before 😬 

Things I still need to implement is supporting the `object` param but I'm not really sure where to start with that one. Please feel free to give me some feedback on this one.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -679,6 +679,14 @@ class Customer(StripeObject):
         return super()._api_delete(id)
 
     @classmethod
+    def _api_retrieve_sources(cls, id):
+        # return 404 if does not exist
+        Customer._api_retrieve(id)
+        
+        obj = cls._api_retrieve(id)
+        return obj.sources
+
+    @classmethod
     def _api_retrieve_source(cls, id, source_id, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
@@ -818,6 +826,9 @@ class Customer(StripeObject):
 
 
 extra_apis.extend((
+    # Retrieve list of sources:
+    ('GET', '/v1/customers/{id}/sources', Customer._api_retrieve_sources),
+    # Add a source
     ('POST', '/v1/customers/{id}/sources', Customer._api_add_source),
     # Retrieve single source by id:
     ('GET', '/v1/customers/{id}/sources/{source_id}',
--- a/test.sh
+++ b/test.sh
@@ -197,6 +197,12 @@ res=$(
   | grep -oE $card)
 [ -n "$res" ]
 
+# observe new card in customer sources response
+res=$(
+  curl -sSf -u $SK: $HOST/v1/customers/$cus/sources \
+  | grep -oE $card)
+[ -n "$res" ]
+
 # delete the card
 curl -sSf -u $SK: $HOST/v1/customers/$cus/sources/$card \
      -X DELETE


meatherly:
What would you suggest for filtering out sources on their type for the `object` type param? Would I just initialize a new `List` then filter out the `obj.sources` based on their prefix? 

something like: (pseudo code)
```
returnList = List('/v1/customers/' + self.id + '/sources')
returnList._list = obj.sources.filter source.startswith(object)
```

meatherly:
Looking at their docs it looks like `object` param can only be `card` or `bank_account`

localstripe/resources.py
```diff
@@ -678,6 +678,11 @@ def _api_delete(cls, id):
         schedule_webhook(Event('customer.deleted', obj))
         return super()._api_delete(id)
 
+    @classmethod
+    def _api_retrieve_sources(cls, id):        
+        obj = cls._api_retrieve(id) # return 404 if does not exist
```
adrienverge, line 683: Does this code really retrieve a list of sources that belong to a customer?

localstripe/resources.py
```diff
@@ -678,6 +678,11 @@ def _api_delete(cls, id):
         schedule_webhook(Event('customer.deleted', obj))
         return super()._api_delete(id)
 
+    @classmethod
+    def _api_retrieve_sources(cls, id):
```
adrienverge, line 682: To be consistent with the rest of the code, maybe `_api_list_sources` would be better.

meatherly:
Man can I just say you're awesome for creating this and for being patient with me as I'm working on this!!

localstripe/resources.py
meatherly: I believe so because the sources are set here:
https://github.com/adrienverge/localstripe/blob/f5a2e0bc4330624741c4e3e655648d5c00f49452/localstripe/resources.py#L629

And then updated here:
https://github.com/adrienverge/localstripe/blob/f5a2e0bc4330624741c4e3e655648d5c00f49452/localstripe/resources.py#L739
and: 
https://github.com/adrienverge/localstripe/blob/f5a2e0bc4330624741c4e3e655648d5c00f49452/localstripe/resources.py#L754

So I'm just returning that value from the customer object

localstripe/resources.py
meatherly: Changed

meatherly:
Alright went off the beaten path did an override on `Card._api_list_all` to allow filtering of customer cards. If you're not a fan of this I can move that logic into the `Customer._api_list_sources`.

Let me know what you think

localstripe/resources.py
```diff
@@ -318,6 +318,22 @@ def _attaching_is_declined(self):
...
+        try:
+            if customer is not None:
+                assert type(customer) is str and customer.startswith('cus_')
+        except AssertionError:
+            raise UserError(400, 'Bad request')
+
+        if customer:
+            Customer._api_retrieve(customer)  # to return 404 if not existant
+
+        li = super(Card, cls)._api_list_all(url, limit=limit)
```
adrienverge, line 332: I think `super()` is enough.

Anyway, if you move this code inside the `Customer` class, you can simply call `Card._api_list_all()` which is even clearer :wink:

localstripe/resources.py
```diff
@@ -678,6 +694,28 @@ def _api_delete(cls, id):
         schedule_webhook(Event('customer.deleted', obj))
         return super()._api_delete(id)
 
+    @classmethod
+    def _api_list_sources(cls, id, object=None, limit=10):
+        print(object)
```
adrienverge, line 699: Please don't include this in final code.

localstripe/resources.py
```diff
@@ -678,6 +694,28 @@ def _api_delete(cls, id):
...
+        obj = cls._api_retrieve(id) # return 404 if does not exist
+        
+        url = '/v1/customers/' + obj.id + '/sources'
+
+        if object is 'card':
+            return Card._api_list_all(url, customer=obj.id)
+
+        #TODO: implement bank accounts
+        if object is 'bank_account':
+            return List(url, limit=limit)
```
adrienverge, line 715: I propose to simplify all this and don't use the `object` argument. Just check it, but don't use it, and return *all* sources. What do you think?

localstripe/resources.py
meatherly: Good catch!

PR #149, opened 2020-09-09
Invoice: Add 'billing_reason' property

adrienverge:
Let's add the property `billing_reason` that was missing. For the
moment, it's not implemented so it's always `null`.

https://stripe.com/docs/api/invoices/object#invoice_object-billing_reason


H--o-l (approved):
Thanks @adrienverge !

PR #154, opened 2020-09-25
Allow email query param in /v1/customers

### LINKED ISSUE #153: /v1/customers doesn't accept email filter query param
[this line](https://github.com/adrienverge/localstripe/blob/38add1cecbd447d777c7cedc803f79302113c1bb/localstripe/resources.py#L166) raises an exception if receiving a GET request to /v1/customers containing [email parameter](https://stripe.com/docs/api/customers/list#list_customers-email) e.g.
`stripe.Customer.list(email = 'address@domain.com')`

lowlyocean:
Resolves #153

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -161,7 +161,7 @@ class StripeObject(object):
         return {"deleted": True, "id": id}
 
     @classmethod
-    def _api_list_all(cls, url, limit=None, **kwargs):
+    def _api_list_all(cls, url, limit=None, email=None, **kwargs):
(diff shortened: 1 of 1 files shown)


adrienverge:
Cool!

The parameter is now accepted, but not handled. What about really processing it?
You can an example on how we filter subscriptions (but are many others examples):
https://github.com/adrienverge/localstripe/blob/e6c5886427cfdb1d762088ca612347dbf73230da/localstripe/resources.py#L2664-L2673

PR #151, opened 2020-09-24
type not required in creating product

lowlyocean:
… return callstack in UserError

Resolves #150

PR #156, opened 2020-09-25
Enable callstack in 400 response, if STRIPE_LOG=debug

### LINKED ISSUE #155: Lack of debugging in case localstripe runs remotely
Would be great if some environment variable could be enabled such that queries from another host resulting in 400 response can also send along the callstack

lowlyocean:
Resolves #155


adrienverge:
Thanks for the proposal.
I don't see the usefulness of such an addition, that would moreover leak information about the host system + Python paths. I believe it was useful for you when you debugged #151? But won't be useful for localstripe users when they *use* the server.
What do you think?

lowlyocean:
Yes I was thinking about the potential to leak, it's a valid concern. However the approach here is to require an environment variable to be purposefully enabled in to activate the behavior. My use case is a Gitlab CI pipeline which spawns localstripe as a Docker container, and then my existing end-to-end tests are run against it. It's quite helpful in the CI logs to see why Stripe requests failed.

Issue #164, opened 2020-12-24
Implement balance object


tarasyarema:
This issue will be to handle the `GET /v1/balance` resource.

tarasyarema:
Hi @H--o-l, I implemented the `/v1/balance` resource. I'm planning on implementing the other resources I noted in the last issue #163. Wanted to know what is the cycle of PR acceptance.

H--o-l:
Hi @tarasyarema ! Thanks ! 
I'm in holiday right now. Will look into it the 4th or 5th January

tarasyarema:
Cool, just wanted to know, have nice holidays!

adrienverge:
Hello @tarasyarema, thanks for your contributions. There is no need to open empty issues to track changes, especially if there are pull requests already. Let's follow this in #165.

PR #165, opened 2020-12-24
Balance resource implementation

tarasyarema:
Implementes #164

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -244,6 +244,53 @@ class StripeObject(object):
         return obj
 
 
+class Balance(object):
+    object = 'balance'
+
+    def __init__(self):
+        self.livemode = False
+        self.available = {
+            'amount': 2000,
+            'currency': 'eur',
+            'source_types': {
+                'card': 2000
+            }
+        }
+        self.pending = {
+            'amount': 0,
+            'currency': 'eur',
+            'source_types': {
+                'card': 0
+            }
+        }
+
+        store[self.object] = self
+
+        schedule_webhook(Event('balance.available', self))
+
+    @classmethod
+    def _api_retrieve(self):
+        obj = store.get(self.object)
+        if obj is None:
+            return self()
+        return obj
+
+    def _export(self, expand=None):
+        obj = {}
+
+        for key, value in vars(self).items():
+            if not key.startswith('_'):
+                if isinstance(value, dict):
+                    obj[key] = value.copy()
+                else:
+                    obj[key] = value
+
+        return obj
+
+
+extra_apis.append(('GET', '/v1/balance', Balance._api_retrieve))
+
+
 class Card(StripeObject):
     object = 'card'
     _id_prefix = 'card_'
--- a/test.sh
+++ b/test.sh
@@ -781,3 +781,5 @@ no_more_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
 zero_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
                  | grep -oE '^  "data": \[\]')
 [ -n "$zero_events" ]
+
+curl -sSfg -u $SK: $HOST/v1/balance


localstripe/resources.py
```diff
@@ -2878,3 +2878,67 @@ def __init__(self, card=None, customer=None, **kwargs):
 
         self.type = 'card'
         self.card = card_obj
+
+
+class Balance(StripeObject):
+    object = 'balance'
+    # Balance does not have id prefix
+    _id_prefix = 'balance_'
```
H--o-l, line 2886: Then it shouldn't b defined ;-)

localstripe/resources.py
```diff
@@ -2878,3 +2878,67 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+            for a in avaliable
+        ]
+
+        pending_obj = [
+            self._get_fund(amount=p["amount"], currency=p["currency"])
+            for p in pending
+        ]
+
+        # All exceptions must be raised before this point.
+        super().__init__(id="")
```
H--o-l, line 2913: I propose:
```
# All exceptions must be raised before this point.
super().__init__(id="balance")
```
I think it make more sense.
Also, it should be move just after 
```
        if kwargs:
        # All exceptions must be raised before this point.
        super().__init__(id="balance")
```
because (I'm not 100% sure) the other objects call it as soon as possible (just after all API arguments been checked)

localstripe/resources.py
```diff
@@ -2878,3 +2878,67 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        # https://stripe.com/docs/api/balance/balance_object
+        avaliable_obj = [
+            self._get_fund(amount=a["amount"], currency=a["currency"])
+            for a in avaliable
+        ]
+
+        pending_obj = [
+            self._get_fund(amount=p["amount"], currency=p["currency"])
+            for p in pending
+        ]
```
H--o-l, line 2910: Two points:
- `avaliable` => `available` (I see the error 4 times in the commit).
- For `avaliable_obj` and `avaliable` I propose to use only one variable, ie `self.availables`, same for "pending".

localstripe/resources.py
```diff
@@ -2878,3 +2878,67 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        super().__init__(id="")
+
+        self.type = 'balance'
+
+        self.avaliable = avaliable_obj
+        self.pending = pending_obj
+
+        schedule_webhook(Event('balance.avaliable', self))
+
+    def _get_fund(self, amount=0, currency="usd"):
```
H--o-l, line 2922: I think default value (`0` and `usd`) are not needed here, and don't help to read the code.

localstripe/resources.py
```diff
@@ -2878,3 +2878,67 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+class Balance(StripeObject):
+    object = 'balance'
+    # Balance does not have id prefix
+    _id_prefix = 'balance_'
+
+    def __init__(self, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        _currencies = ["usd", "eur"]
```
H--o-l, line 2892: You should use only `"` or only `'` not a mix of `'` and `"`.
In localstripe we use `'` as much as possible.
Can you change all `"` into `'`?

H--o-l (commented):
Thanks for your work!

Interesting Stripe object this [`balance`](https://stripe.com/docs/api/balance/balance_object), I think it's the first one, and unique, I see that exist only once. All other resources can exist multiple times.

But maybe one day other resources like this would exist, and I also think it's worth to be aware of this possibility in the `serve.py` file.

Thus, above your existing commits I propose the following diff:
```diff
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -146,6 +146,13 @@ class StripeObject(object):
 
         return obj
 
+    @classmethod
+    def _api_get(cls):
+        obj = store.get(cls.object, None)
+        if obj is None:
+            return cls()
+        return obj
+
     @classmethod
     def _api_update(cls, id, **data):
         obj = cls._api_retrieve(id)
@@ -2927,18 +2934,3 @@ class Balance(StripeObject):
                 "card": amount
             }
         }
-
-    @classmethod
-    def _api_get(cls, id=""):
-        # Try getting the balance from the store, and if
-        # if its the first time, we create a new one
-        try:
-            obj = cls._api_retrieve(id)
-        except UserError:
-            obj = Balance()
-
-        return obj
-
-
-extra_apis.append((
-    ('GET', '/v1/balance', Balance._api_get)))
--- a/localstripe/server.py
+++ b/localstripe/server.py
@@ -24,7 +24,7 @@ import socket
 
 from aiohttp import web
 
-from .resources import Charge, Coupon, Customer, \
+from .resources import Balance, Charge, Coupon, Customer,\
                        Event, Invoice, InvoiceItem, PaymentIntent, \
                        PaymentMethod, Plan, Product, Refund, SetupIntent, \
                        Source, Subscription, SubscriptionItem, TaxRate, \
@@ -214,6 +214,15 @@ def api_retrieve(cls, url):
     return f
 
 
+def api_get(cls, url):
+    async def f(request):
+        data = await get_post_data(request) or {}
+        data.update(unflatten_data(request.query) or {})
+        expand = data.pop('expand', None)
+        return json_response(cls._api_get()._export(expand=expand))
+    return f
+
+
 def api_update(cls, url):
     async def f(request):
         id = request.match_info['id']
@@ -275,6 +284,10 @@ for cls in (Charge, Coupon, Customer, Event, Invoice, InvoiceItem,
             ('GET', '/v1/' + cls.object + 's', api_list_all)):
         app.router.add_route(method, url, func(cls, url))
 
+for cls in (Balance, ):
+    for method, url, func in (('GET', '/v1/' + cls.object, api_get), ):
+        app.router.add_route(method, url, func(cls, url))
+
 
 def localstripe_js(request):
     path = os.path.dirname(os.path.realpath(__file__)) + '/localstripe-v3.js'
```
I quickly tested it and I think it works, but don't hesitate to change it if I missed something.

Last things, it's perfect to have splitter change in multiple Pull requests.
But inside this PR I think all three commits should be only one.
Reasons are:
- we usually merge unit test inside the commit that adds the corresponding feature
- the "style" commit can be merge too inside the commit that add the feature, because this commit is still editable (the commit is not merged yet). For readability of the PR, and the final Git branch, it will be better

localstripe/resources.py
tarasyarema: Actually in the Payouts PR I found it it does, its `ba_`, adding it.

tarasyarema:
Really good approach @H--o-l with the `StripeObject` modifications! As this was my first PR I had no deep knowledge of this class. Adding it in the following commit. Also I'll squash all commits and force push a single commit with `feat: added Balance resource`.

H--o-l (commented):
Neat, I find the new diff way simpler, thanks @tarasyarema.

Playing with the new API I realize that there was an unexpected field returned:
```
> curl -sSfg -u sk_test_12345: "localhost:8420/v1/balance"
{
  "available": [
    {
      "amount": 0,
      "currency": "usd",
      "source_types": {
        "card": 0
      }
    },
    {
      "amount": 0,
      "currency": "eur",
      "source_types": {
        "card": 0
      }
    }
  ],
  "created": 1609856080,
  "id": "balance",  <== should not be here
  "livemode": false,
  "object": "balance",
  "pending": [
    {
      "amount": 0,
      "currency": "usd",
      "source_types": {
        "card": 0
      }
    },
    {
      "amount": 0,
      "currency": "eur",
      "source_types": {
        "card": 0
      }
    }
  ],
  "type": "balance"
}
```
I think it revealed an issue with this id.
Thus, for the `balance` object, I propose to create an object store in the `Store` like the other but without the `id` field, using only the class name.
I propose the following diff:
```diff
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -111,8 +111,7 @@ class StripeObject(object):
 
     def __init__(self, id=None):
         if not isinstance(self, List):
-            if id is None:
-                assert hasattr(self, '_id_prefix')
+            if id is None and hasattr(self, '_id_prefix'):
                 self.id = getattr(self, '_id_prefix') + random_id(14)
             else:
                 self.id = id
@@ -121,7 +120,10 @@ class StripeObject(object):
 
             self.livemode = False
 
-            key = self.object + ':' + self.id
+            if self.id:
+                key = self.object + ':' + self.id
+            else:
+                key = self.object
             if key in store.keys():
                 raise UserError(409, 'Conflict')
             store[key] = self
@@ -2889,14 +2891,13 @@ class Token(StripeObject):
 
 class Balance(StripeObject):
     object = 'balance'
-    _id_prefix = ''
 
     def __init__(self, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
         # All exceptions must be raised before this point.
-        super().__init__(id='balance')
+        super().__init__()
 
         _currencies = ['usd', 'eur']
 ```
With it the API answer is:
```
> curl -sSfg -u sk_test_12345: "localhost:8420/v1/balance"
{
  "available": [
    {
      "amount": 0,
      "currency": "usd",
      "source_types": {
        "card": 0
      }
    },
    {
      "amount": 0,
      "currency": "eur",
      "source_types": {
        "card": 0
      }
    }
  ],
  "created": 1609856080,
  "id": null,
  "livemode": false,
  "object": "balance",
  "pending": [
    {
      "amount": 0,
      "currency": "usd",
      "source_types": {
        "card": 0
      }
    },
    {
      "amount": 0,
      "currency": "eur",
      "source_types": {
        "card": 0
      }
    }
  ],
  "type": "balance"
}
```
It's still not perfect, but I think it's closer to what the real Stripe API return (for info [the documentation example](https://stripe.com/docs/api/balance/balance_object) said that Stripe doesn't return an `id`).
@adrienverge will have to validate all this when he is available, but for me it's a good approach.

**Review by H--o-l — APPROVED** [review]
@tarasyarema LGTM, thanks!
By warmed it's possible that @adrienverge have a completely different review but :crossed_fingers: 
I will look at #167 Monday, sorry for the delay.

**Comment by tarasyarema** [comment]
No problem, thanks!

File: localstripe/resources.py [hunk 6]
```diff
@@ -2878,3 +2887,41 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+    object = 'balance'
+
+    def __init__(self, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        # All exceptions must be raised before this point.
+        super().__init__()
+
+        currencies = ['usd', 'eur']
```
adrienverge, line 2902: This list is arbitrary, and `usd` was never used in localstripe before. Do you need it, or can we keep `eur` for consistency?

localstripe/resources.py
```diff
@@ -2878,3 +2887,41 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        schedule_webhook(Event('balance.available', self))
+
+    def _get_fund(self, amount, currency):
+        return {
+            'amount': amount,
+            'currency': currency,
+            'source_types': {
+                'card': amount
+            }
+        }
```
adrienverge, line 2927: I think it would be more readable to drop this function and just use:
```python
        self.pending = [{'amount': 0,
                         'currency': currency,
                         'source_types': {'card': 0}}
                        for currency in currencies]
```

adrienverge (commented):
Thanks @tarasyarema for your work and @H--o-l for the review :+1: 

I think `Balance` objects should not inherit from `StripeObject`:
- they would have methods like `_api_create()`, `_api_update()`, `_api_delete()`, `_api_list_all()`, `_get_class_for_id()`, etc., which doesn't make sense,
- they would have properties like `created`, etc., which doesn't make sense,
- it requires `StripeObject`s to be modified (to support not having `id`s for instance).

Instead, they `Balance` could be regular objects (just inherit from `object`) and save their state in the global `store` object.

What do you think?

localstripe/server.py
```diff
@@ -275,6 +284,10 @@ async def f(request):
             ('GET', '/v1/' + cls.object + 's', api_list_all)):
         app.router.add_route(method, url, func(cls, url))
```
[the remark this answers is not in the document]
tarasyarema: Sorry, actually I tend do what my linter tells me... thanks for the noticing!

tarasyarema:
As you can see, @adrienverge , I made some changes related to what you said:
1. The object is not inherited from `StripeObject`, thus is defines its own `_api_get` and `_export` [1] and the path is handled by `api_extra` so it does not generate any overhead to the server.
2. Clean up the code a bit and added a few fields that are needed for the `_export` method.

[1] Note that to be coherent with the server api the export method need to handle a positional argument `expand`, which I just set to `None`. I did not want to go deeper into this issue, as maybe it could add additional complexity to server, so I think is fine, the logic is simple anyways,

localstripe/resources.py
```diff
@@ -244,6 +246,76 @@ def do_expand(path, obj):
...
+                    obj[key] = value._export()
+                elif (isinstance(value, list) and len(value) and
+                        isinstance(value[0], StripeObject)):
+                    obj[key] = [item._export() for item in value]
+                elif isinstance(value, dict):
+                    obj[key] = value.copy()
+                else:
+                    obj[key] = value
+
+        return obj
```
H--o-l, line 312: I think that what Adrien was proposing here https://github.com/adrienverge/localstripe/pull/165#discussion_r555092875 is to use a `Balance._api_retrieve()` direcly in `server.py`, not `api_retrieve()`. Like that no need to define an `_export()` method.
Do you think it's possible?

localstripe/resources.py
```diff
@@ -121,7 +120,10 @@ def __init__(self, id=None):
 
             self.livemode = False
 
-            key = self.object + ':' + self.id
+            if self.id:
+                key = self.object + ':' + self.id
+            else:
+                key = self.object
```
H--o-l, line 126: This change and the one above (that I made you do sorry about that) are not needed anymore with the last commit version, no?

localstripe/resources.py
```diff
@@ -244,6 +246,76 @@ def do_expand(path, obj):
...
+        self.connect_reserved = []
+        self.available = [
+            {
+                'amount': 2000,
+                'currency': currency,
+                'source_types': {
+                    'card': 2000
+                }
+            }
+            for currency in currencies
```
H--o-l, line 270: If there is only one currency maybe this part can be simplified?
A PR is perfect not when there is nothing to add, but when there is nothing to remove :P

localstripe/resources.py
```diff
@@ -244,6 +246,76 @@ def do_expand(path, obj):
...
+                elif isinstance(value, dict):
+                    obj[key] = value.copy()
+                else:
+                    obj[key] = value
+
+        return obj
+
+
+extra_apis.append((
+    ('GET', '/v1/balance', Balance._api_get)))
```
H--o-l, line 316: Can it be kept inside `server.py`? 
If think it's what you were proposing @adrienverge but I'm not sure.

adrienverge: I'm OK if it stays here, as long as it's consistent. If it helps reducing the changes needed in `server.py`, why not.

Style notes:
- there are extra parentheses,
- it fits on one line.

localstripe/resources.py
adrienverge: Yes, let's not duplicate code by redefining `_export()` (unless there is really no other way). And we use `_api_retrieve()` (not `_api_get()`) because `GET` are associated to multiple operations (including *retrieve*).

tarasyarema: Yes, but remember that now it inherits from just `object`... so there are no methods

tarasyarema: The problem is with the webhook schedule:

Screenshot 2021-01-13 at 11 00 52

tarasyarema: So we actually _need_ to have the `_export` method defined

localstripe/server.py
```diff
@@ -25,10 +25,10 @@
...
 
 from .resources import Charge, Coupon, Customer, \
-                       Event, Invoice, InvoiceItem, PaymentIntent, \
-                       PaymentMethod, Plan, Product, Refund, SetupIntent, \
-                       Source, Subscription, SubscriptionItem, TaxRate, \
-                       Token, extra_apis, store
+    Event, Invoice, InvoiceItem, PaymentIntent, \
+    PaymentMethod, Plan, Product, Refund, SetupIntent, \
+    Source, Subscription, SubscriptionItem, TaxRate, \
+    Token, extra_apis, store
```
H--o-l, line 31: I think this PR doesn't add or remove imports anymore.
If it's true, it shouldn't modify these lines.

localstripe/resources.py
H--o-l: I agree that, because `extra_apis.append((})` call `func(**data)._export(expand=expand)`, we can't have both:
- untouched `server.py`
- no `Balance._export()` defined

But I think it's possible to have either:
- a simpler `Balance._export()` by stripping all the code that is unnecessary for the `Balance` object.
- or adding a few lines of code inside `server.py`, like:
```py
for cls in (Balance, ):
    for method, url, func in (('GET', '/v1/' + cls.object, cls._api_retrieve), ):
        app.router.add_route(method, url, func(cls, url))
```
With `cls._api_retrieve` defined inside `ressource.py`.

Both solutions work for me (personally I would have tried the second one first).

localstripe/server.py
tarasyarema: Sorry, the linter :/

localstripe/resources.py
```diff
@@ -244,6 +244,67 @@ def do_expand(path, obj):
         return obj
 
 
+class Balance(object):
+    object = 'balance'
+
+    def __init__(self, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
```
adrienverge, line 252: If this object cannot be created nor update (no `PUT` nor `POST`), then `__init__()` will never be called with keyword arguments. You can remove it and drop lines 251-252.

localstripe/resources.py
```diff
@@ -244,6 +244,67 @@ def do_expand(path, obj):
...
+
+        self.object = self.object
```
adrienverge, line 254: What does this mean? If it has no use, please drop it.

localstripe/resources.py
```diff
@@ -244,6 +244,67 @@ def do_expand(path, obj):
...
+        self.livemode = False
+        self.type = 'balance'
```
adrienverge, line 256: I do not see any `type` on the Balance object: https://stripe.com/docs/api/balance/balance_object

localstripe/resources.py
```diff
@@ -244,6 +244,67 @@ def do_expand(path, obj):
...
+        self.connect_reserved = []
```
adrienverge, line 257: If it's not implemented, the best is probably not to include `connect_reserved`. Would you be OK to remove it?

localstripe/resources.py
```diff
@@ -244,6 +244,67 @@ def do_expand(path, obj):
...
+                'card': 0
+            }
+        }
+
+        store[self.object] = self
+
+        schedule_webhook(Event('balance.available', self))
+
+    @classmethod
+    def _api_get(self):
```
adrienverge, line 278: [Again](https://github.com/adrienverge/localstripe/pull/165#discussion_r556375194), the correct term is `_api_retrieve`.

localstripe/resources.py
```diff
@@ -244,6 +244,67 @@ def do_expand(path, obj):
...
+        obj = store.get(self.object, None)
```
adrienverge, line 279: `, None` is useless, please drop it.

localstripe/resources.py
```diff
@@ -244,6 +244,67 @@ def do_expand(path, obj):
...
+        if obj is None:
+            return self()
+        return obj
+
+    def _export(self, expand=None):
+        # expand is passed to be coherent with the server api
+        # but it's not used
```
adrienverge, line 286: If it's not implemented, please drop it: it will make potentiel errors clearer.
Moreover, it seems to make no sense to pass `expand` to a Balance request.

localstripe/resources.py
```diff
@@ -244,6 +244,67 @@ def do_expand(path, obj):
...
+
+        obj = {}
+
+        for key, value in vars(self).items():
+            if not key.startswith('_'):
+                if isinstance(value, StripeObject):
+                    obj[key] = value._export()
+                elif (isinstance(value, list) and len(value) and
+                        isinstance(value[0], StripeObject)):
+                    obj[key] = [item._export() for item in value]
```
adrienverge, line 296: All these lines are useless, aren't they? Can they be removed?

localstripe/resources.py
tarasyarema: Yes.

localstripe/resources.py
tarasyarema: Oh, ok sorry I did not get what you meant

tarasyarema:
About [this comment](https://github.com/adrienverge/localstripe/pull/165#discussion_r561110081) its actually needed as I see.

Screenshot 2021-01-21 at 13 02 08

H--o-l:
Hi @tarasyarema!
There are two commits inside this PR now, can you remove the commit called "Merge branch 'master' into balance"?
(I think it's the same for the payouts PR).
Thanks!

tarasyarema:
Oh, my bad. I though as its a merge with master the squash was not needed. I'll squash this tomorrow morning and when it gets merged I squash the merge in Payouts (?).

H--o-l:
No Issue!
I'm not sure a squash will work.
The issue is that you made a "merge" with master, while for a PR, I think it's better to do a "rebase".
Now you can:
- ensure the master of your fork is up to date https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork (but I think you already did this)
- `git reset @~1 --hard` to remove the `Merge branch 'master' into balance` commit
- `git rebase origin/master` to rebase the `balance` branch on top of the master branch.
- and push

tarasyarema:
@H--o-l done! Thanks for the commands :D

adrienverge (approved):
Thanks for your contribution @tarasyarema!

And thanks @H--o-l  for your review :+1:

Issue #166, opened 2020-12-29
Implement Payouts resource


tarasyarema:
This issue will be to handle the `/v1/payouts` resource.

adrienverge:
Hello @tarasyarema, thanks for your contributions. There is no need to open empty issues to track changes, especially if there are pull requests already. Let's follow this in #167!

PR #167, opened 2020-12-29
Payouts implementation

tarasyarema:
This implements #166

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -2097,6 +2097,103 @@ class Plan(StripeObject):
         return Product._api_retrieve(self.product).statement_descriptor
 
 
+class Payout(StripeObject):
+    object = 'payout'
+    _id_prefix = 'po_'
+
+    def __init__(self, amount=None, currency=None, description=None,
+                 metadata=None, statement_descriptor=None, destination=None,
+                 method=None, source_type=None, status=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        amount = try_convert_to_int(amount)
+        try:
+            assert type(amount) is int and amount > 0
+            assert currency in ('eur',)
+            if description is not None:
+                assert type(description) is str
+            if metadata is not None:
+                assert type(metadata) is dict
+            if statement_descriptor is not None:
+                assert type(statement_descriptor) is str \
+                    and len(statement_descriptor) <= 22
+            if method is not None:
+                assert method in ('standard', 'instant')
+            if source_type is not None:
+                assert type(source_type) is str
+            if status is not None:
+                assert status in ('paid', 'pending', 'failed')
+        except AssertionError:
+            raise UserError(400, 'Bad request')
+
+        # All exceptions must be raised before this point.
+        super().__init__()
+
+        self.amount = amount
+        self.currency = currency
+        self.description = description or ''
+        self.destination = f'ba_{random_id(24)}'
+        self.metadata = metadata or {}
+        self.method = method or 'standard'
+        self.source_type = source_type or 'card'
+        self.statement_descriptor = statement_descriptor or ''
+
+        two_days = 60 * 60 * 24 * 2
+        self.arrival_date = int(time.time() + two_days)
+
+        # Payout scheduling is not implemented yet so all payouts are
+        # manually created
+        self.automatic = False
+        # Balance Transactions are no implemented yet so we fake one
+        self.balance_transaction = f"txn_{random_id(24)}"
+
+        self.failure_balance_transaction = None
+        self.failure_code = None
+        self.failure_message = None
+        self.original_payout = None
+        self.reversed_by = None
+        self.status = status or 'pending'
+        self.type = 'bank_account'
+
+        schedule_webhook(Event('payout.created', self))
+
+        if status == 'failed':
+            self.failure_balance_transaction = ''
+            self.failure_code = 'could_not_process'
+            self.failure_message = 'The bank could not process this payout.'
+
+        if status in ('paid', 'failed'):
+            schedule_webhook(Event(f'payout.{status}', self))
+
+    @classmethod
+    def _api_update(cls, id, **data):
+        obj = super()._api_update(id, **data)
+        schedule_webhook(Event('payout.updated', obj))
+        return obj
+
+    @classmethod
+    def _api_cancel(cls, id, **kwargs):
+        payout = Payout._api_retrieve(id)
+
+        # Only pending payouts can be canceled
+        if payout.status != 'pending':
+            raise UserError(400, 'Cannot cancel payout')
+
+        payout._update(status='canceled')
+
+        schedule_webhook(Event('payout.canceled', payout))
+
+        return payout
+
+    @classmethod
+    def _api_delete(cls, id):
+        raise UserError(405, 'Method Not Allowed')
+
+
+extra_apis.append(('POST', '/v1/payouts/{id}/cancel', Payout._api_cancel))
+
+
 class Product(StripeObject):
     object = 'product'
     _id_prefix = 'prod_'
--- a/localstripe/server.py
+++ b/localstripe/server.py
@@ -24,11 +24,10 @@ import socket
 
 from aiohttp import web
 
-from .resources import Charge, Coupon, Customer, \
-                       Event, Invoice, InvoiceItem, PaymentIntent, \
-                       PaymentMethod, Plan, Product, Refund, SetupIntent, \
-                       Source, Subscription, SubscriptionItem, TaxRate, \
-                       Token, extra_apis, store
+from .resources import Charge, Coupon, Customer, Event, Invoice, \
+    InvoiceItem, PaymentIntent, PaymentMethod, Payout, Plan, Product, Refund, \
+    SetupIntent, Source, Subscription, SubscriptionItem, TaxRate, Token, \
+    extra_apis, store
 from .errors import UserError
 from .webhooks import register_webhook
 
@@ -265,8 +264,9 @@ for method, url, func in extra_apis:
 
 
 for cls in (Charge, Coupon, Customer, Event, Invoice, InvoiceItem,
-            PaymentIntent, PaymentMethod, Plan, Product, Refund, SetupIntent,
-            Source, Subscription, SubscriptionItem, TaxRate, Token):
+            PaymentIntent, PaymentMethod, Payout, Plan, Product, Refund,
+            SetupIntent, Source, Subscription, SubscriptionItem, TaxRate,
+            Token):
     for method, url, func in (
             ('POST', '/v1/' + cls.object + 's', api_create),
             ('GET', '/v1/' + cls.object + 's/{id}', api_retrieve),
--- a/test.sh
+++ b/test.sh
@@ -783,3 +783,31 @@ zero_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
 [ -n "$zero_events" ]
 
 curl -sSfg -u $SK: $HOST/v1/balance
+
+payout=$(
+  curl -sSfg -u $SK: $HOST/v1/payouts \
+       -d amount=1100 \
+       -d currency=eur \
+  | grep -oE 'po_\w+' | head -n 1)
+
+payout_status=$(
+  curl -sSfg -u $SK: $HOST/v1/payouts/$payout \
+  | grep -oE '"status": "pending",')
+[ -n "$payout_status" ]
+
+curl -sg -u $SK: $HOST/v1/payouts/$payout/cancel -X POST
+
+payout_status=$(
+  curl -sSfg -u $SK: $HOST/v1/payouts/$payout \
+  | grep -oE '"status": "canceled",')
+[ -n "$payout_status" ]
+
+curl -sg -u $SK: $HOST/v1/payouts \
+      -d amount=1100 \
+      -d currency=eur \
+      -d status=paid
+
+curl -sg -u $SK: $HOST/v1/payouts \
+      -d amount=1100 \
+      -d currency=eur \
+      -d status=failed


localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        self.method = method
+        self.source_type = source_type
+        self.statement_descriptor = statement_descriptor
+
+        # By default expected arrival to two days after creation
+        two_days = 60 * 60 * 24 * 2
+        self.arrival_date = int(time.time() + two_days)
+
+        # As it was manually created
+        self.automatic = False
```
H--o-l, line 2994: At payout creation this property is "True" by default no?
I think I don't understand the comment entirely.

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        self.balance_transaction = f"txn_{random_id(24)}"
```
H--o-l, line 2995: OK, maybe here comment is necessary, like "balance_transaction don't exist in localstripe yet, fake one" (or something better if you have an idea).

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+
+        # Cancel related fields
```
H--o-l, line 2997: Not needed I think

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        self.failure_balance_transaction = None
+        self.failure_code = None
+        self.failure_message = None
+
+        # No payout reverse logic
```
H--o-l, line 3002: We don't add comment for un-implemented feature.
There would be a lot of comment otherwise I think.

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        self.original_payout = None
+        self.reversed_by = None
+
+        # Default status on creation
+        self.status = 'pending'
+
+        # Default
+        self.type = "bank_account"
```
H--o-l, line 3010: All these default values, this one and the one above, can you remove empty lines between them?

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        # Cancel related fields
+        self.failure_balance_transaction = None
+        self.failure_code = None
+        self.failure_message = None
+
+        # No payout reverse logic
+        self.original_payout = None
+        self.reversed_by = None
+
+        # Default status on creation
```
H--o-l, line 3006: not needed

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        self.status = 'pending'
+
+        # Default
```
H--o-l, line 3009: not needed ^^

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        self.type = "bank_account"
+
+        schedule_webhook(Event('payout.created', self))
+
+        if status in ('paid', 'failed'):
+            self.status = status
```
H--o-l, line 3015: when the if is false this line should not be executed?
That sounds weird.

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+
+            # Define some arbitrary failure fields
+            # based on https://stripe.com/docs/api/payouts/failures
```
H--o-l, line 3018: Hum, I'd be for removing this comment too, I think the code is clear enought.

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+
+        return payout
+
+    @classmethod
+    def _api_cancel(cls, id):
+        payout = Payout._api_retrieve(id)
+
+        # Only pending payouts can be canceled
+        if payout.status != 'pending':
+            raise UserError(400, 'Could not cancel not pending payout')
```
H--o-l, line 3042: I read that only "manual" payout can be canceled, maybe we can add this check?
In that case, I would use the same answer message for both, something like "Cannot cancel payout".

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+        data = {"status": "canceled"}
+        payout._update(**data)
+
+        schedule_webhook(Event('payout.canceled', payout))
+
+        return payout
+
+    @classmethod
+    def _api_delete(cls, id):
+        raise UserError(405, 'Method Not Allowed')
```
H--o-l, line 3053: It's not your fault but I don't understand why Stripe use a `POST payout/cancel` route instead of using `DELETE payout` :sob: .
Anyway, great implementation :+1:

H--o-l (commented):
Hi @tarasyarema, thanks for your work!

Please see my comments below.
Also, can you merge the three commits into one commit?

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
 
         self.type = 'card'
         self.card = card_obj
+
+
+class Payout(StripeObject):
```
adrienverge, line 2883: Please keep `StripeObject` subclasses alphabetically sorted.

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+    _id_prefix = 'po_'
+
+    def __init__(self, amount=None, currency=None, description=None,
+                 metadata=None, statement_descriptor=None, destination=None,
+                 method=None, source_type=None, no_funds=None,
+                 status=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        try:
```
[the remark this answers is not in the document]
adrienverge: In addition to @H--o-l's comment, can you also check out what is done in other classes, and try to replicate it here? For example:
- all `try_convert_to_...()` are done before the `try` block,
- one-liners like `assert type(amount) is int and amount > 0` are used,
- `UserError` are *never* raised inside this `try` block, only assertions are used,
- etc.

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+            amount = try_convert_to_int(amount)
+            assert type(amount) is int
+            assert amount > 0
+
+            assert type(currency) is str
+
+            # TODO
+            # Support all Stripe currencies https://stripe.com/docs/currencies
+            if currency not in ('eur', 'usd'):
+                raise UserError(400, f'Currency {currency} not supported')
```
adrienverge, line 2904: Why?

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+            if metadata is not None:
+                assert type(metadata) is dict
+            else:
+                metadata = {}
+
+            if statement_descriptor is not None:
+                assert type(statement_descriptor) is str
+
+                if len(statement_descriptor) > 22:
+                    msg = 'Parameter "statement_desciptor" '\
```
adrienverge, line 2921: desciptor → descriptor

localstripe/resources.py
```diff
@@ -2878,3 +2878,180 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+    @classmethod
+    def _api_cancel(cls, id):
+        payout = Payout._api_retrieve(id)
+
+        # Only pending payouts can be canceled
+        if payout.status != 'pending':
+            raise UserError(400, 'Could not cancel not pending payout')
+
+        data = {"status": "canceled"}
+        payout._update(**data)
```
adrienverge, line 3045: This is equivalent to
```suggestion
        payout._update(status='canceled')
```

localstripe/server.py
```diff
@@ -25,10 +25,10 @@
...
 
 from .resources import Charge, Coupon, Customer, \
-                       Event, Invoice, InvoiceItem, PaymentIntent, \
-                       PaymentMethod, Plan, Product, Refund, SetupIntent, \
-                       Source, Subscription, SubscriptionItem, TaxRate, \
-                       Token, extra_apis, store
+    Event, Invoice, InvoiceItem, PaymentIntent, \
+    PaymentMethod, Plan, Product, Refund, SetupIntent, \
+    Source, Subscription, SubscriptionItem, TaxRate, \
+    Token, Payout, extra_apis, store
```
adrienverge, line 31: Same for alphabet (and same below).

localstripe/resources.py
tarasyarema: Yes, actually its `False`, Stripe doing their thing... https://stripe.com/docs/api/payouts/object#payout_object-automatic we put it to false because '_false if it was requested manually._'

localstripe/resources.py
tarasyarema: There are only few Payout event types: https://stripe.com/docs/api/events/types#event_types-payout.canceled

tarasyarema: Thus, yes, in the other cases we do not care about this `if`

tarasyarema:
@adrienverge and @H--o-l thank you very much for this (and the other PRs) reviews! implemented the comments and sorry for consistently not following the style... Hope to get more used to it soon. 
Thanks for the amazing in project and hope everything is ok, I'll have a look at any comments you may have about these final refactors and squashes.

localstripe/resources.py
```diff
@@ -2019,6 +2019,110 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+        self.source_type = source_type or 'card'
+        self.statement_descriptor = statement_descriptor or ''
+
+        two_days = 60 * 60 * 24 * 2
+        self.arrival_date = int(time.time() + two_days)
+
+        # As it was manually created as balance_transaction
+        # doesn't exist in localstripe yet, so we fake one
+        self.automatic = False
+        self.balance_transaction = f"txn_{random_id(24)}"
```
H--o-l, line 2073: OK, I think I start to understand, thus I propose:
```py
# In localstripe, there is no "Payout schedule" yet, all payouts are manually created
self.automatic = False
# In localstripe, there is no "Balance Transaction" yet, so we fake one
self.balance_transaction = f"txn_{random_id(24)}"
```
If I missed something don't hesitate to tell, I never played with these route on the real Stripe.

localstripe/resources.py
```diff
@@ -2019,6 +2019,110 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+
+        self.failure_balance_transaction = None
+        self.failure_code = None
+        self.failure_message = None
+        self.original_payout = None
+        self.reversed_by = None
+        self.status = 'pending'
```
H--o-l, line 2080: Maybe the following?
```
self.status = status or 'pending'
```
I think it simplified the code, it's easier to use only one variable `self.status` than two variables `self.status` and `status`. And it's more consistent with the rest of the code, no?

localstripe/resources.py
```diff
@@ -2019,6 +2019,110 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+    _id_prefix = 'po_'
+
+    def __init__(self, amount=None, currency=None, description=None,
+                 metadata=None, statement_descriptor=None, destination=None,
+                 method=None, source_type=None, status=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        amount = try_convert_to_int(amount)
+
```
H--o-l, line 2033: I see in the rest of the code that there is no empty line here usually.
Could you remove it?

localstripe/resources.py
```diff
@@ -2019,6 +2019,110 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+        self.original_payout = None
+        self.reversed_by = None
+        self.status = 'pending'
+        self.type = 'bank_account'
+
+        schedule_webhook(Event('payout.created', self))
+
+        if status in ('paid', 'failed'):
+            self.status = status
+            if status == 'failed':
```
H--o-l, line 2087: If you follow my previous comment, here you can simplify the two `if` and keep only one.
Oh but now I see the `schedule_webhook(Event(f'payout.{status}', self))`.
I think `schedule_webhook(Event(f'payout.{status}', self))` would be better in his own dedicated `if`, splitter things like that help reading the code (even if variable are tested multiple time, that is not an issue).

H--o-l (commented):
> @adrienverge and @H--o-l thank you very much for this (and the other PRs) reviews! implemented the comments and sorry for consistently not following the style... Hope to get more used to it soon.

No issue, it's hard to get inside a project!

localstripe/resources.py
tarasyarema: Cool, no problem!

localstripe/resources.py
tarasyarema: So you mean not to interpolate?

localstripe/resources.py
```diff
@@ -2019,6 +2019,110 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+                 metadata=None, statement_descriptor=None, destination=None,
+                 method=None, source_type=None, status=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        amount = try_convert_to_int(amount)
+
+        try:
+            assert type(amount) is int and amount > 0
+            assert type(currency) is str and currency in ('eur')
```
adrienverge, line 2036: Beware, in Python, `('eur')` *is* the string `'eur'`. If you want to declare a list, use `('eur', )`.

However, is there a need to enforce only some currencies?

localstripe/resources.py
H--o-l: What I had in mind is the following:
```py
if status in 'failed':
    self.failure_balance_transaction = ''	
    self.failure_code = 'could_not_process'	
    self.failure_message = \	         
        'The bank could not process this payout.'	    

if status in ('paid', 'failed'):
    schedule_webhook(Event(f'payout.{status}', self))
```
You see it's simpler because the logic of settings the "failure" fields is differentiated from the logic of sending webhook.
You can see them as two different features.
They share a common condition that activates them sure, but the most important is that they are two different features.
When the code shows it clearly, it becomes easier to read it (at least I think).

tarasyarema:
@H--o-l the changes you said!

H--o-l (approved):
Looks great for me, thanks @tarasyarema !

localstripe/resources.py
```diff
@@ -2050,6 +2050,111 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+    def __init__(self, amount=None, currency=None, description=None,
+                 metadata=None, statement_descriptor=None, destination=None,
+                 method=None, source_type=None, status=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        amount = try_convert_to_int(amount)
+        try:
+            assert type(amount) is int and amount > 0
+            assert type(currency) is str and currency in ('eur',)
```
adrienverge, line 2066: This can be simplified to:
```suggestion
            assert currency in ('eur',)
```

localstripe/resources.py
```diff
@@ -2050,6 +2050,111 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+            if description is not None:
+                assert type(description) is str
+            if metadata is not None:
+                assert type(metadata) is dict
+            if statement_descriptor is not None:
+                assert type(statement_descriptor) is str \
+                    and len(statement_descriptor) <= 22
+            if method is not None:
+                assert type(method) is str \
+                    and method in ('standard', 'instant')
```
adrienverge, line 2076: This can be simplified to:

```suggestion
                assert method in ('standard', 'instant')
```
(same for `status` below)

localstripe/server.py
```diff
@@ -265,8 +265,9 @@ async def f(request):
 
 
 for cls in (Charge, Coupon, Customer, Event, Invoice, InvoiceItem,
-            PaymentIntent, PaymentMethod, Plan, Product, Refund, SetupIntent,
-            Source, Subscription, SubscriptionItem, TaxRate, Token):
+            PaymentIntent, PaymentMethod, Payout, Plan, Product, Refund,
+            SetupIntent, Source, Subscription, SubscriptionItem,
+            TaxRate, Token,):
```
adrienverge, line 270: To keep the same coding style:
```suggestion
            SetupIntent, Source, Subscription, SubscriptionItem, TaxRate,
            Token):
```

localstripe/server.py
```diff
@@ -25,10 +25,10 @@
...
 
 from .resources import Charge, Coupon, Customer, \
-                       Event, Invoice, InvoiceItem, PaymentIntent, \
-                       PaymentMethod, Plan, Product, Refund, SetupIntent, \
-                       Source, Subscription, SubscriptionItem, TaxRate, \
-                       Token, extra_apis, store
+    Event, Invoice, InvoiceItem, PaymentIntent, \
+    PaymentMethod, Payout, Plan, Product, Refund, SetupIntent, \
+    Source, Subscription, SubscriptionItem, TaxRate, \
+    Token, extra_apis, store
```
adrienverge, line 31: I'm not opposed to changing indentation, but in this case please fill up lines up to 79 chars.

localstripe/resources.py
```diff
@@ -2050,6 +2050,111 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+
+        schedule_webhook(Event('payout.created', self))
+
+        if status == 'failed':
+            self.failure_balance_transaction = ''
+            self.failure_code = 'could_not_process'
+            self.failure_message = \
+                'The bank could not process this payout.'
+
+        if status in ('paid', 'failed',):
```
adrienverge, line 2123: No need for a trailing comma, to be consistent let's use:
```suggestion
        if status in ('paid', 'failed'):
```

localstripe/resources.py
```diff
@@ -2050,6 +2050,111 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+        self.status = status or 'pending'
+        self.type = 'bank_account'
+
+        schedule_webhook(Event('payout.created', self))
+
+        if status == 'failed':
+            self.failure_balance_transaction = ''
+            self.failure_code = 'could_not_process'
+            self.failure_message = \
+                'The bank could not process this payout.'
```
adrienverge, line 2121: This looks better:
```suggestion
            self.failure_message = 'The bank could not process this payout.'
```

localstripe/resources.py
```diff
@@ -2050,6 +2050,111 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+        self.metadata = metadata or {}
+        self.method = method or 'standard'
+        self.source_type = source_type or 'card'
+        self.statement_descriptor = statement_descriptor or ''
+
+        two_days = 60 * 60 * 24 * 2
+        self.arrival_date = int(time.time() + two_days)
+
+        # Payout scheduling is not implemented yet
+        # so all payouts are manually created
```
adrienverge, line 2101: Here and everywhere in the PR, can you please fill up lines up to 79 characters, so it's easier to read?

localstripe/resources.py
```diff
@@ -2050,6 +2050,111 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+            self.failure_message = \
+                'The bank could not process this payout.'
+
+        if status in ('paid', 'failed',):
+            schedule_webhook(Event(f'payout.{status}', self))
+
+    @classmethod
+    def _api_update(cls, id, **data):
+        payout = Payout._api_retrieve(id)
+        payout._update(**data)
```
adrienverge, line 2129: It's better not to duplicate code (for multiple reasons), so let's use the parent class code instead of rewriting it.

For example see how it's done in `Customer`:
https://github.com/adrienverge/localstripe/blob/e6a6af2/localstripe/resources.py#L720-L722

localstripe/resources.py
```diff
@@ -2050,6 +2050,111 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+
+        schedule_webhook(Event('payout.updated', payout))
+
+        return payout
+
+    @classmethod
+    def _api_cancel(cls, id):
```
adrienverge, line 2136: Can you please use this? Passing an argument to the route would return a clear error instead of crashing the server loop.
```suggestion
    def _api_cancel(cls, id, **kwargs):
        if kwargs:
            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
```

localstripe/resources.py
```diff
@@ -2050,6 +2050,111 @@ def statement_descriptor(self):  # Support Stripe API <= 2018-02-05
...
+
+        return payout
+
+    @classmethod
+    def _api_delete(cls, id):
+        raise UserError(405, 'Method Not Allowed')
+
+
+extra_apis.append((
+    ('POST', '/v1/payouts/{id}/cancel', Payout._api_cancel)))
```
adrienverge, line 2155: Again, there seems to be 2 extra parentheses here.

test.sh
```diff
@@ -781,3 +781,52 @@ no_more_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
...
+
+# check correct status
+payout_status=$(
+  curl -sSfg -u $SK: $HOST/v1/payouts/$payout \
+  | grep -oE '"status": "pending",')
+[ -n "$payout_status" ]
+
+# cancel the payout
+curl -sSfg -u $SK: $HOST/v1/payouts/$payout/cancel \
+  -X POST
```
adrienverge, line 800: You don't want the `-S` and `-f` options here. Especiall `-f`, which is used to fail silently on server errors.

Better to use:
```suggestion
curl -sg -u $SK: $HOST/v1/payouts/$payout/cancel -X POST
```

test.sh
```diff
@@ -781,3 +781,52 @@ no_more_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
 zero_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
                  | grep -oE '^  "data": \[\]')
 [ -n "$zero_events" ]
+
+# create a pending payout
```
adrienverge, line 785: Would you be OK to remove comments? Code looks self explanatory to me (here and all comments below).

test.sh
```diff
@@ -781,3 +781,52 @@ no_more_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
...
+  curl -sSfg -u $SK: $HOST/v1/payouts \
+       -d amount=1100 \
+       -d currency=eur \
+       -d status=failed \
+  | grep -oE 'po_\w+' | head -n 1)
+
+payout_status=$(
+  curl -sSfg -u $SK: $HOST/v1/payouts/$payout \
+  | grep -oE '"status": "failed",')
+[ -n "$payout_status" ]
```
adrienverge, line 832: The first test "create payout, check correct status, cancel it, check different status" makes sense. But in the 2 other ones, it's not very useful to create an object and check it's property `status` just after.

I propose to remove them, because we don't have such basic tests for other classes, and it makes tests a bit heavier to read.

If you *really* want to keep them, they can at least be simplified:
```sh
curl -sg -u $SK: $HOST/v1/payouts \
      -d amount=1100 \
      -d currency=eur \
      -d status=paid

curl -sg -u $SK: $HOST/v1/payouts \
      -d amount=1100 \
      -d currency=eur \
      -d status=failed
```

adrienverge (commented):
Thanks @tarasyarema, the logic looks good, just a few things to fix.

adrienverge (approved):
Thanks @tarasyarema!

Issue #168, opened 2020-12-30
Implement pagination


tarasyarema:
This is kind of a bigger issue that arises when working with Stripe packages. The packages use a pre-defined pagination structure [1]. 

The idea is to implement the following query parameters in the `List` type, so those can be passed and used for paginations.

- ~`limit`~ already implemented
- `starting_after`
- `ending_before`

[1] Stripe docs: Pagination. https://stripe.com/docs/api/pagination

PR #169, opened 2020-12-30
List pagination: starting_after

tarasyarema:
implements `starting_after` query param for `List` as noted in #168

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -161,11 +161,11 @@ class StripeObject(object):
         return {"deleted": True, "id": id}
 
     @classmethod
-    def _api_list_all(cls, url, limit=None, **kwargs):
+    def _api_list_all(cls, url, limit=None, starting_after=None, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
-        li = List(url, limit=limit)
+        li = List(url, limit=limit, starting_after=starting_after)
         li._list = [value for key, value in store.items()
                     if key.startswith(cls.object + ':')]
         return li
@@ -476,7 +476,8 @@ class Charge(StripeObject):
         return self.amount <= self.amount_refunded
 
     @classmethod
-    def _api_list_all(cls, url, customer=None, created=None, limit=10):
+    def _api_list_all(cls, url, customer=None, created=None, limit=10,
+                      starting_after=None):
         try:
             if customer is not None:
                 assert type(customer) is str and customer.startswith('cus_')
@@ -499,7 +500,8 @@ class Charge(StripeObject):
             if type(created) is str or not created.get('gt'):
                 raise UserError(500, 'Not implemented')
 
-        li = super(Charge, cls)._api_list_all(url, limit=limit)
+        li = super(Charge, cls)._api_list_all(url, limit=limit,
+                                              starting_after=starting_after)
         if customer:
             li._list = [c for c in li._list if c.customer == customer]
         if created and created.get('gt'):
@@ -1254,7 +1256,8 @@ class Invoice(StripeObject):
         return super()._api_delete(id)
 
     @classmethod
-    def _api_list_all(cls, url, customer=None, subscription=None, limit=None):
+    def _api_list_all(cls, url, customer=None, subscription=None, limit=None,
+                      starting_after=None):
         try:
             if customer is not None:
                 assert type(customer) is str and customer.startswith('cus_')
@@ -1264,7 +1267,8 @@ class Invoice(StripeObject):
         except AssertionError:
             raise UserError(400, 'Bad request')
 
-        li = super(Invoice, cls)._api_list_all(url, limit=limit)
+        li = super(Invoice, cls)._api_list_all(url, limit=limit,
+                                               starting_after=starting_after)
         if customer is not None:
             Customer._api_retrieve(customer)  # to return 404 if not existant
             li._list = [i for i in li._list if i.customer == customer]
@@ -1435,14 +1439,17 @@ class InvoiceItem(StripeObject):
         self.metadata = metadata or {}
 
     @classmethod
-    def _api_list_all(cls, url, customer=None, limit=None):
+    def _api_list_all(cls, url, customer=None, limit=None,
+                      starting_after=None):
         try:
             if customer is not None:
                 assert type(customer) is str and customer.startswith('cus_')
         except AssertionError:
             raise UserError(400, 'Bad request')
 
-        li = super(InvoiceItem, cls)._api_list_all(url, limit=limit)
+        li = super(InvoiceItem,
+                   cls)._api_list_all(url, limit=limit,
+                                      starting_after=starting_after)
         li._list = [ii for ii in li._list if ii.invoice is None]
         if customer is not None:
             Customer._api_retrieve(customer)  # to return 404 if not existant
@@ -1513,11 +1520,13 @@ class InvoiceLineItem(StripeObject):
 class List(StripeObject):
     object = 'list'
 
-    def __init__(self, url=None, limit=None):
+    def __init__(self, url=None, limit=None, starting_after=None):
         limit = try_convert_to_int(limit)
         limit = 10 if limit is None else limit
         try:
             assert type(limit) is int and limit > 0
+            if starting_after is not None:
+                assert type(starting_after) is str and len(starting_after) > 0
         except AssertionError:
             raise UserError(400, 'Bad request')
 
@@ -1527,11 +1536,16 @@ class List(StripeObject):
         self.url = url
 
         self._limit = limit
+        self._starting_after = starting_after
+        self._starting_pos = None
         self._list = []
 
     @property
     def data(self):
-        return [item._export() for item in self._list][:self._limit]
+        self._compute_starting_pos()
+        return [item._export() for item in self._list[
+            self._starting_pos:self._starting_pos + self._limit
+        ]]
 
     @property
     def total_count(self):
@@ -1539,7 +1553,21 @@ class List(StripeObject):
 
     @property
     def has_more(self):
-        return len(self._list) > self._limit
+        self._compute_starting_pos()
+        return len(self._list) > self._limit + self._starting_pos
+
+    def _compute_starting_pos(self):
+        if self._starting_pos is not None:
+            return
+
+        self._starting_pos = 0
+        if self._starting_after is None:
+            return
+
+        for i, item in enumerate(self._list):
+            if getattr(item, 'id', None) == self._starting_after:
+                self._starting_pos = i + 1
+                break
 
 
 class PaymentIntent(StripeObject):
@@ -1911,7 +1939,8 @@ class PaymentMethod(StripeObject):
         return super()._api_retrieve(id)
 
     @classmethod
-    def _api_list_all(cls, url, customer=None, type=None, limit=None):
+    def _api_list_all(cls, url, customer=None, type=None, limit=None,
+                      starting_after=None):
         try:
             assert _type(customer) is str and customer.startswith('cus_')
             assert type in ('card', )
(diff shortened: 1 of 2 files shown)


localstripe/resources.py
```diff
@@ -1513,11 +1513,14 @@ def _api_delete(cls, id):
 class List(StripeObject):
     object = 'list'
 
-    def __init__(self, url=None, limit=None):
+    def __init__(self, url=None, limit=None, starting_after=None):
         limit = try_convert_to_int(limit)
         limit = 10 if limit is None else limit
         try:
             assert type(limit) is int and limit > 0
+
```
adrienverge, line 1521: Please drop this line, for consistency with the rest of the code.

localstripe/resources.py
```diff
@@ -1527,19 +1530,30 @@ def __init__(self, url=None, limit=None):
         self.url = url
 
         self._limit = limit
+        self._starting_after = starting_after
+        self._starting_pos = 0
```
adrienverge, line 1534: This variable looks inconsistent: if `has_more()` is called before `data()`, it will have a wrong value.

test.sh
```diff
@@ -6,6 +6,8 @@ set -eux
 HOST=http://localhost:8420
 SK=sk_test_12345
 
+LIMIT=10000
+
```
adrienverge, line 10: There is no need to define a global variable here, please set it where it's used.

test.sh
```diff
@@ -769,3 +771,14 @@ total_count=$(
...
+# check starting_after parameter for the events
+
+# get the last event id
+events=$(curl -sSfg -u $SK: $HOST/v1/events?limit=$LIMIT)
+event_id=$(echo "$events" | grep -oE 'evt_\w+' | tail -n 1)
+
+# after the last event id there should not be more events
+events_none=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$event_id | grep -oE '"has_more": false')
+[ -n "$events_none" ]
+
```
adrienverge, line 784: Please drop the useless newline at the end.

localstripe/resources.py
```diff
@@ -1527,19 +1530,30 @@ def __init__(self, url=None, limit=None):
...
+        if self._starting_after is None:
+            return [item._export() for item in self._list][:self._limit]
+
+        for i, item in enumerate(self._list):
+            if item._export().get('id', None) == self._starting_after:
+                self._starting_pos = i + 1
+                break
+
+        return [item._export()
+                for item in self._list[self._starting_pos:]][:self._limit]
```
adrienverge, line 1548: The code would be more efficient with
```suggestion
        return [item._export()
                for item in self._list[self._starting_pos:self._starting_pos + self._limit]]
```

localstripe/resources.py
```diff
@@ -1527,19 +1530,30 @@ def __init__(self, url=None, limit=None):
...
         self._list = []
 
     @property
     def data(self):
-        return [item._export() for item in self._list][:self._limit]
+        if self._starting_after is None:
+            return [item._export() for item in self._list][:self._limit]
+
+        for i, item in enumerate(self._list):
+            if item._export().get('id', None) == self._starting_after:
```
adrienverge, line 1543: - Using `.get('id', None)` is equivalent to `.get('id')`.
- Calling `._export()` is resource-consuming: do you think there is a way to use something like `getattr(item, 'id', None)`?

adrienverge (commented):
Thanks for implementing this @tarasyarema. The current approach has a small problem (see following comments), so what about initializing the class with `self._starting_pos = None`; if it's unset (either in `data()` or `has_more()`), compute its value; then simply return the correct value from `self._list[self._starting_pos:self._starting_pos + self._limit]`?

tarasyarema:
Thanks for your comments! Good point with the call of `has_more` before `data`. I implemented a helper method `_compute_starting_pos` that computes the staring position before `has_more` and `data`. Plus, the way it fetches the list is with the `getattr` function as you mentioned. I put it this way because in the code there are some time that arbitrary elements are appended to the list after the init, so we need to check for the starting position every time (if given).

tarasyarema:
Made a simple refactor. 
The value `self._starting_pos` is not really used as its dynamically computed, so it's transferred to the method `self._starting_pos()`.

adrienverge:
Thanks, this new version looks good!

> Made a simple refactor.

I think it was better before this commit `refactor: simplification`, because when exported to JSON, computing the start index (which can be a heavy operation) is done twice, at least. With the previous version, it was done only once.
Beware that the function name `_compute_staring_pos()` has a typo (missing `t`).

Can you also squash all commits into one with a consistent commit message?

tarasyarema:
Done @adrienverge !

tarasyarema:
Btw, don't know why the tests failed only for Python 3.8.

test.sh
```diff
@@ -769,3 +769,11 @@ total_count=$(
   curl -sSfg -u $SK: $HOST/v1/charges?customer=$cus\&created[gt]=1588166306 \
   | grep -oE '"total_count": 6')
 [ -n "$total_count" ]
+
+# check starting_after parameter for the events
+# get the last event id
```
adrienverge, line 774: You can drop the comments, the command are self-explanatory.

test.sh
```diff
@@ -769,3 +769,11 @@ total_count=$(
...
+event_id=$(curl -sSfg -u $SK: $HOST/v1/events?limit=10000 | grep -oE 'evt_\w+' | tail -n 1)
```
adrienverge, line 775: I think the maximum allowed `limit` value is `100`: https://stripe.com/docs/api/pagination#pagination-limit. If we enforce it in the future, this test will fail. Let's use `100`?

test.sh
```diff
@@ -769,3 +769,11 @@ total_count=$(
...
+
+# after the last event id there should not be more events
+events_none=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$event_id | grep -oE '"has_more": false')
+[ -n "$events_none" ]
```
adrienverge, line 779: This test makes sense only if the list is non empty without `starting_after`.

What do you think of this?
```sh
no_more_events=$(curl -sSfg -u $SK: $HOST/v1/events | grep -oE '"has_more": false')
[ -z "$no_more_events" ]
last_event=$(curl -sSfg -u $SK: $HOST/v1/events?limit=100 | grep -oE 'evt_\w+' | tail -n 1)
no_more_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event | grep -oE '"has_more": false')
[ -n "$no_more_events" ]
```

adrienverge (commented):
Thanks again @tarasyarema, but this new version does not do what we talked about: the heavy computation is still done twice or more (every time `_compute_starting_pos()` is called). Why not initializing `self._starting_pos = None`, and compute it only if it's still `None`, inside `_compute_starting_pos()`?

> Btw, don't know why the tests failed only for Python 3.8.

I just restarted the job and it passed.

tarasyarema:
I don't really get what you mean with the computing start... The function is invoked two times, but once the value is set we do not go in the loop, hence the heavy computation is only done once (?)

adrienverge:
@tarasyarema you can add a debugging line to see whether the loop is entered only once, or more:
```diff
@@ -1556,6 +1556,7 @@ class List(StripeObject):
             return
 
         # Only traverse the list when starting_after is given
+        print('entering heavy computation...', flush=True)
         for i, item in enumerate(self._list):
             if getattr(item, 'id', None) == self._starting_after:
                 self._starting_pos = i + 1
```

tarasyarema:
Oh sorry, I forgot about the `_starting_after`... Lemme change it

test.sh
tarasyarema: I don't really get what you doing here.

tarasyarema: The thing is that the nested event object may have a `has_more` field.

localstripe/resources.py
```diff
@@ -161,11 +161,11 @@ def _api_delete(cls, id):
         return {"deleted": True, "id": id}
 
     @classmethod
-    def _api_list_all(cls, url, limit=None, **kwargs):
+    def _api_list_all(cls, url, limit=None, starting_after=None, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
-        li = List(url, limit=limit)
+        li = List(url, limit=limit, starting_after=starting_after)
```
adrienverge, line 168: There are other `_api_list_all()` functions where the same needs to be done (in `Charge`, `Invoice`, `InvoiceItem`, `PaymentMethod`, `Refund` and `Subscription`).
Can you add them there too? You just need to take the argument and pass it.

test.sh
adrienverge: > I don't really get what you doing here.

Here is the idea:
\- List all events, check that there are more (`"has_more": false` is not present).
\- List all events *after the last one*, check that there are no more (`"has_more": false` is present).
An extra check is to verify that zero events are returned on the second call (see my example below).

> The thing is that the nested event object may have a `has_more` field.

That's true. So, this means that the initial test you wrote doesn't check anything neither, right?
In this case, you just need to check `has_more` fields that are correctly indented: `grep -oE '^  "has_more": false'`.

Please use this:
```sh
no_more_events=$(curl -sSfg -u $SK: $HOST/v1/events \
                 | grep -oE '^  "has_more": false' || true)
[ -z "$no_more_events" ]
last_event=$(curl -sSfg -u $SK: $HOST/v1/events?limit=100 \
             | grep -oE 'evt_\w+' | tail -n 1)
no_more_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
                 | grep -oE '^  "has_more": false')
[ -n "$no_more_events" ]
zero_events=$(curl -sSfg -u $SK: $HOST/v1/events?starting_after=$last_event \
                 | grep -oE '^  "data": \[\]')
[ -n "$zero_events" ]
```

adrienverge (commented):
@tarasyarema, you re-requested my review but nothing has changed: the "heavy computation" is still calles 2+ times. What you need to do is [in my previous comment](https://github.com/adrienverge/localstripe/pull/169#pullrequestreview-566978738): initialize `self._starting_pos = None`, and compute it only if it's still `None`, inside `_compute_starting_pos()`.

Can you please fix it and request a review after re-reading the PR twice and manually checking that everything is 100% perfect?

localstripe/resources.py
```diff
@@ -1527,19 +1529,35 @@ def __init__(self, url=None, limit=None):
...
+        return len(self._list) > self._limit + self._starting_pos
+
+    def _compute_starting_pos(self):
+        if self._starting_after is None:
+            return
+
+        for i, item in enumerate(self._list):
+            if getattr(item, 'id', None) == self._starting_after:
+                self._starting_pos = i + 1
+                self._starting_after = None
```
tarasyarema, line 1559: @adrienverge this is the line I changed, so it only does the "heavy computation" once

tarasyarema:
Tested the different resources with `starting_after` parameter and it works only computing one time the `_starting_pos`.

adrienverge:
@tarasyarema I asked you to double checked before I have to review yet another time, and there are still multiple problems:
- I tried your code a few hours ago and it did make 2 "heavy computations" per HTTP call.
- Again: please initialize `self._starting_pos = None` inside `__init__()`, and compute it only if it's still None, inside `_compute_starting_pos()`. Do *not* reassign `self._starting_after`.
- You didn't remove the comment in `test.sh` (again).

tarasyarema:
1. I literally put a print like you mentioned [in this comment](https://github.com/adrienverge/localstripe/pull/169#issuecomment-759302467) and saw only one comment per request, but anyways... Implemented the way you mentioned it.
2. Sorry about the comment, forgot to remove them.

adrienverge (approved):
Thanks

tarasyarema:
Thanks for merging! About the docker image, will it be updated too?

adrienverge:
It will on the next release; probably after #165 and #167 and reviewed and merged.

PR #174, opened 2021-02-03
BalanceTransactions: implement BalanceTransaction resource

brandondyck:


bravier (commented):
Hi @brandondyck,
Thanks for your contribution on localstripe!
PR looks nice and I have a few comment, please see below ↓

localstripe/resources.py
```diff
@@ -291,6 +291,78 @@ def _export(self, expand=None):
...
+            assert type in ('charge', 'refund')
+        except AssertionError:
+            raise UserError(400, 'Bad request')
+
+        if source.startswith('ch_'):
+            Charge._api_retrieve(source)
+        elif source.startswith('re_'):
+            Refund._api_retrieve(source)
+        else:
+            raise UserError(400, 'Bad request')
```
adrienverge, line 321: Can you add these comments? Otherwise it's not obvious for new developers why these lines are for.
```suggestion
        if source.startswith('ch_'):
            Charge._api_retrieve(source)  # to return 404 if not existant
        elif source.startswith('re_'):
            Refund._api_retrieve(source)  # to return 404 if not existant
        else:
            raise UserError(400, 'Bad request')
```

localstripe/resources.py
```diff
@@ -291,6 +291,78 @@ def _export(self, expand=None):
...
+        self.status = 'available'
+        self.type = type
+
+    @property
+    def net(self):
+        return self.amount - self.fee
+
+    @classmethod
+    def _api_list_all(cls, url, limit=None, starting_after=None, **kwargs):
+        li = super(BalanceTransaction, cls)._api_list_all(url, limit, starting_after, **kwargs)
```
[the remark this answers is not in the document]
adrienverge: In addition to @bravier's comment, I think we can use `super()` instead of `super(BalanceTransaction, cls)`. Along with removing `**kwargs`, it should solve the line length problem.

localstripe/resources.py
```diff
@@ -291,6 +291,78 @@ def _export(self, expand=None):
...
+        li._list.sort(key=lambda b: b.created, reverse=True)
```
adrienverge, line 345: Is it really needed? If not, we could remove this new `_api_list_all()` function and let the parent class do the job?

adrienverge (commented):
Three tiny comments/questions in addition to @bravier's review (thanks!)

PS: In the commit title: `implemented` → `implement`

localstripe/resources.py
bravier: I think it's needed as it's what the doc says on listing _Balance Transactions_:
> The transactions are returned in sorted order, with the most recent transactions appearing first.

https://stripe.com/docs/api/balance_transactions/list

adrienverge: @bravier OK thanks!

localstripe/resources.py
```diff
@@ -291,6 +291,78 @@ def _export(self, expand=None):
...
+        # All exceptions must be raised before this point
+        super().__init__()
+
+        self.amount = amount
+        self.available_on = self.created
+        self.currency = currency
+        self.description = description
+        self.exchange_rate = exchange_rate
+        self.fee = 0
+        self.fee_details = []
```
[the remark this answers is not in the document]
brandondyck: Yes. The docs aren't clear, but the API always gives me `[]` instead of `null` for `fee_details`.

localstripe/resources.py
```diff
@@ -291,6 +291,78 @@ def _export(self, expand=None):
...
+        self.source = source
+        self.status = 'available'
+        self.type = type
+
+    @property
+    def net(self):
+        return self.amount - self.fee
+
+    @classmethod
+    def _api_list_all(cls, url, limit=None, starting_after=None, **kwargs):
```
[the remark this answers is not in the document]
brandondyck: I removed `**kwargs` from the `super` call, but I left it in the signature because otherwise I get a warning that the signature doesn't match the base method, and because I can use it to check for unsupported params.

localstripe/resources.py
```diff
@@ -291,6 +291,78 @@ def _export(self, expand=None):
...
+
+    def __init__(self, amount=None, currency=None, description=None,
+                 exchange_rate=None, reporting_category=None, source=None,
+                 type=None,
+                 **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        try:
+            assert _type(amount) is int
```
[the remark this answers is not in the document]
brandondyck: Done for both `amount` and `exchange_rate`.

brandondyck:
I've addressed your comments, including fixing the commit title.

bravier (approved):
LGTM! Thanks @brandondyck!

localstripe/resources.py
```diff
@@ -2280,6 +2371,18 @@ def __init__(self, charge=None, amount=None, metadata=None, **kwargs):
...
+    @classmethod
+    def _api_create(cls, **data):
+        obj = super()._api_create(**data)
+        txn = BalanceTransaction(amount=-obj.amount,
+                                 currency=obj.currency,
+                                 description='REFUND FOR CHARGE',
+                                 exchange_rate=1.0,
+                                 reporting_category='refund',
+                                 source=obj.id, type='refund')
+        obj.balance_transaction = txn.id
```
adrienverge, line 2383: Usually the creation of related entities is rather done inside `__init__()`. Is it possible to do the same here?

It could go inside a block `if self.status == 'succeeded'`, for clarity (even if it's the only refund status implemented for now).

adrienverge (commented):
Thanks @brandondyck. I have a small remark, please see below.

localstripe/resources.py
brandondyck: Done.

adrienverge (approved):
Thanks @brandondyck!

PR #175, opened 2021-02-03
localstripe.js: Make compatible with react-stripe-js

### LINKED ISSUE #147: Support for latest stripe-js and react-stripe-js
`react-stripe-js` is doing a validation on the stripe object: https://github.com/stripe/react-stripe-js/blob/e13fae556b1a667a5dfa1130c00d5ec4c2bc154e/src/utils/guards.ts#L16

This validation is failing because of the missing `createPaymentMethod`

brandondyck:
* Add stubbed createPaymentMethod
* Add getElement to elements in JS client
* Fix typo in createToken
* Add unmount and destroy to Element

This fixes #147.

### DIFF
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -86,7 +86,7 @@ class Element {
 
   mount(domElement) {
     if (typeof domElement === 'string') {
-      domElement = document.querySelector(domElement)[0];
+      domElement = document.querySelector(domElement);
     }
 
     const span = document.createElement('span');
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -357,6 +357,8 @@ Stripe = (apiKey) => {
         }
       }
     },
+
+    createPaymentMethod: async () => ({}),
   };
 };
 
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -80,20 +80,40 @@ function openModal(text, confirmText, cancelText) {
 }
 
 class Element {
-  constructor() {
+  constructor(stripeElements) {
+    // Element needs a reference to the object that created it, in order to
+    // thoroughly destroy() itself.
+    this._stripeElements = stripeElements;
     this.listeners = {};
+    this._domChildren = [];
   }
 
   mount(domElement) {
     if (typeof domElement === 'string') {
       domElement = document.querySelector(domElement);
+    } else if (!(domElement instanceof window.Element)) {
+      throw new Error('Invalid DOM element. Make sure to call mount() with ' +
+                      'a valid DOM element or selector.');
     }
 
-    const span = document.createElement('span');
-    span.textContent = 'localstripe: ';
-    domElement.appendChild(span);
+    if (this._stripeElements._cardElement !== this) {
+      throw new Error('This Element has already been destroyed. Please ' +
+                      'create a new one.');
+    }
 
-    const inputs = {
+    if (this._domChildren.length) {
+      if (domElement === this._domChildren[0].parentElement) {
+        return;
+      }
+      throw new Error('This Element is already mounted. Use `unmount()` to ' +
+                      'unmount the Element before re-mounting.');
+    }
+
+    const labelSpan = document.createElement('span');
+    labelSpan.textContent = 'localstripe: ';
+    this._domChildren.push(labelSpan);
+
+    this._inputs = {
       number: null,
       exp_month: null,
       exp_year: null,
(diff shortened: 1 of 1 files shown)


localstripe/localstripe-v3.js
```diff
@@ -151,17 +164,25 @@ Stripe = (apiKey) => {
   return {
     elements: () => {
       return {
-        create: (type, options) => {
+        create: function(type, options) {
```
adrienverge, line 167: Why going back to the 2nd millennium? :wink: 
(by the way you add an arrow function on line 382).

localstripe/localstripe-v3.js
```diff
@@ -151,17 +164,25 @@ Stripe = (apiKey) => {
...
           console.log('localstripe: Stripe().elements().create()');
-          return new Element();
+          if (this.cardElement) {
+            throw new Error("Can only create one Element of type card");
+          }
+          this.cardElement = new Element(this);
+          return this.cardElement;
         },
+        getElement: function(type) {
```
adrienverge, line 175:
```suggestion
        getElement: type => {
```

localstripe/localstripe-v3.js
```diff
@@ -151,17 +164,25 @@ Stripe = (apiKey) => {
...
+          console.log('localstripe: Stripe().elements().getElement()');
```
adrienverge, line 176: Is this log line very valuable? If not, I'd prefer not spamming the console and remove it.

adrienverge (commented):
Thanks for contributing!

This doesn't look bad but I don't understand all the changes. Can you please split each change in an individual commit (easy ones could even be merged in separated PRs, if you'd like), and explain the motivation in each commit message, so reviewers can understand their goal?

For example, I don't get why you add `this.elements` and `this.inputs` to the `Element` class. With real Stripe, they don't seem to exist:
```js
const stripe = Stripe('pk_...');
const element = stripe.elements().create('card');
element.elements  // ← undefined
element.inputs    // ← undefined
```

localstripe/localstripe-v3.js
brandondyck: Because I need to access `this.cardElement`. [According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), an arrow function "Does not have its own bindings to `this` or `super`, and should not be used as methods."

localstripe/localstripe-v3.js
brandondyck: Same reason as for `elements.create`.

localstripe/localstripe-v3.js
brandondyck: It doesn't matter to me. I was just copying what I saw elsewhere, so I'll take it out.

brandondyck:
> Can you please split each change in an individual commit (easy ones could even be merged in separated PRs, if you'd like), and explain the motivation in each commit message, so reviewers can understand their goal?

Sure, I will split it up and clarify my intent. 🙂

brandondyck:
I've split this up into several commits. The last one is still rather complex, but I don't think there's a reasonable way to avaid that. 🤷 

I also made more changes to fix some problems I found while addressing your comments:
1. I found a bug in the way `mount` was using `document.querySelector`.
2. The `'localstripe:'` span was not being removed by `unmount`.
3. Unexpected sequences of `mount`, `unmount`, and `destroy` calls could leave the elements in an inconsistent state and did not match the behavior of Stripe.js very well, so I introduced a field in `Element` to explicitly track mounting/destruction state. This was kind of a large change, and I'm sorry to give you more work to do after you've already kindly reviewed my PR. 😅  I also hope you don't mind the long essay I wrote in the commit message. 😁

localstripe/localstripe-v3.js
```diff
@@ -357,6 +357,8 @@ Stripe = (apiKey) => {
         }
       }
     },
+
+    createPaymentMethod: async () => ({}),
```
adrienverge, line 434:
```suggestion
    createPaymentMethod: async () => {},
```

localstripe/localstripe-v3.js
adrienverge: > Because I need to access `this.cardElement`.

OK!

> You can blame the JavaScript spec for being inconsistent.

;)

localstripe/localstripe-v3.js
```diff
@@ -80,20 +80,49 @@ function openModal(text, confirmText, cancelText) {
...
+  constructor(elements) {
     this.listeners = {};
+    // Element needs a reference to the object that created it, in order to
+    // thoroughly destroy() itself.
+    this._elements = elements;
   }
 
+  // _state tracks calls to unmount, mount, and destroy, and is always one
+  // of 'unmounted', 'mounted', or 'destroyed'.
+  _state = 'unmounted';
```
adrienverge, line 92: 1. This would probably be better inside the constructor, along with other initializations?
2. It's a bit more complex to maintain 3 states represented as string, and I don't feel `destroyed` + warnings "This Element has already been destroyed..." are very useful.
   A simpler solution would be 2 states with something like `this._mounted` (boolean):
   ```js
     constructor(elements) {
       this.listeners = {};
       // Element needs a reference to the object that created it, in order to
       // thoroughly destroy() itself.
       this._elements = elements;
       this._mounted = false;
     }
   ```
   But in my opinion the best would just be an array `this._htmlChildren = []` (described in my review comment).

localstripe/localstripe-v3.js
```diff
@@ -80,20 +80,49 @@ function openModal(text, confirmText, cancelText) {
 }
 
 class Element {
-  constructor() {
+  constructor(elements) {
     this.listeners = {};
+    // Element needs a reference to the object that created it, in order to
+    // thoroughly destroy() itself.
+    this._elements = elements;
```
adrienverge, line 87: - Much better with a leading `_` for these internal properties :+1: 
- What about `parent` instead of `elements`? It's a bit clearer to me, but this is personal taste. Take the proposal or don't.

localstripe/localstripe-v3.js
```diff
@@ -80,20 +80,49 @@ function openModal(text, confirmText, cancelText) {
...
+  // _state tracks calls to unmount, mount, and destroy, and is always one
+  // of 'unmounted', 'mounted', or 'destroyed'.
+  _state = 'unmounted';
+
   mount(domElement) {
     if (typeof domElement === 'string') {
       domElement = document.querySelector(domElement);
+    } else if (!(domElement instanceof window.Element)) {
+      throw new Error('Invalid DOM element. Make sure to call mount() with ' +
+          'a valid DOM element or selector.');
```
adrienverge, line 99: It's just visual style, but the rest of the file uses this kind of indentation:
```suggestion
      throw new Error('Invalid DOM element. Make sure to call mount() with ' +
                      'a valid DOM element or selector.');
```

adrienverge (commented):
Thanks! The commit messages are crystal-clear :ok_hand: 

The 4 first commits are good to go.

About the last one, but I find it a bit complex for what it's doing. Especially maintaining a state (mounted/unmounted) with many inter-related variables: `this._state`, `this._span`, `this._inputs`.
Do you think you could just have one: `this._htmlChildren = []`?
\- if empty → Element is not mounted
\- if has items (span + inputs) → Element is mounted
(and in my opinion we don't have to care about the Element being destroyed twice)

localstripe/localstripe-v3.js
brandondyck: I think `stripeElements` is clearer than either one. The naming is a little tricky because we are working with both DOM elements and Stripe elements, and the DOM has its own concept of "parent".

brandondyck:
I've simplified `Element` using @adrienverge's idea of keeping an array of children. `Element` no longer matches Stripe's behavior as closely. `mount` still throws an exception if the element has been destroyed, because I think it's important to prevent the broken state of having a duplicate set of input fields with no access to it via `getElement`.

localstripe/localstripe-v3.js
adrienverge: `this._stripeElements` sound perfect to me :+1:

adrienverge (approved):
Neat. Thanks for your contribution @brandondyck!

PR #176, opened 2021-02-03
Charge: add statement_descriptor property

brandondyck:
This does not attempt to replicate Stripe's handling of non-ASCII characters statement_descriptor.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -372,6 +372,7 @@ class Charge(StripeObject):
 
     def __init__(self, amount=None, currency=None, description=None,
                  metadata=None, customer=None, source=None, capture=True,
+                 statement_descriptor=None,
                  **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
@@ -390,6 +391,10 @@ class Charge(StripeObject):
                 assert (source.startswith('pm_') or source.startswith('src_')
                         or source.startswith('card_'))
             assert type(capture) is bool
+            if statement_descriptor is not None:
+                assert type(statement_descriptor) is str
+                assert len(statement_descriptor) <= 22
+                assert re.search('[a-zA-Z]', statement_descriptor)
         except AssertionError:
             raise UserError(400, 'Bad request')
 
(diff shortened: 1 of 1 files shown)


localstripe/resources.py
```diff
@@ -390,6 +391,10 @@ def __init__(self, amount=None, currency=None, description=None,
                 assert (source.startswith('pm_') or source.startswith('src_')
                         or source.startswith('card_'))
             assert type(capture) is bool
+            if statement_descriptor is not None:
+                assert type(statement_descriptor) is str
+                assert len(statement_descriptor) <= 22
+                assert re.search('[a-zA-Z]', statement_descriptor)
```
adrienverge, line 397: - Do you want to use `re.match()` or `re.search()`?
  Beware that the following command succeeds:
  ```python
  >>> re.search('[a-zA-Z]', ' A B C   ×€¼¥   ')
  <re.Match object; span=(1, 2), match='A'>
  ```
- Where does the regex `[a-zA-Z]` come from? Why not `[0-9]` too? I couldn't find info in Stripe docs.

adrienverge (commented):
Thanks for contributing @brandondyck!

The change looks good. Just see my three comments below.

localstripe/resources.py
brandondyck: `re.search()` is correct. It is only ensuring that the descriptor contains at least one letter, as the docs specify. My experiments show that neither digits nor letters with precomposed diacritics count toward this requirement.

localstripe/resources.py
```diff
@@ -390,6 +391,10 @@ def __init__(self, amount=None, currency=None, description=None,
                 assert (source.startswith('pm_') or source.startswith('src_')
                         or source.startswith('card_'))
             assert type(capture) is bool
+            if statement_descriptor is not None:
+                assert type(statement_descriptor) is str
+                assert len(statement_descriptor) <= 22
```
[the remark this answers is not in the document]
brandondyck: The regex on the next line ensures that there is at least one letter, so `1 <=` would be redundant.

localstripe/resources.py
adrienverge: Ah OK! I initially thought you wanted to check "only latin characters in the string", not "at least one".

Makes sense then :+1:

localstripe/resources.py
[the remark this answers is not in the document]
adrienverge: :+1:

adrienverge (approved):
Thanks @brandondyck!

PR #183, opened 2021-06-28
Add payment_method to customer creation endpoint

meshy:
Stripe's customer creation endpoint accepts a `payment_method` parameter. This adds the parameter to the localstripe mock.

Ref: https://stripe.com/docs/api/customers/create#create_customer-payment_method

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -710,7 +710,7 @@ class Customer(StripeObject):
                  phone=None, address=None,
                  invoice_settings=None, business_vat_id=None,
                  preferred_locales=None, tax_id_data=None,
-                 metadata=None, **kwargs):
+                 metadata=None, payment_method=None, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
@@ -751,8 +751,13 @@ class Customer(StripeObject):
                 assert set(data.keys()) == {'type', 'value'}
                 assert data['type'] in ('eu_vat', 'nz_gst', 'au_abn')
                 assert type(data['value']) is str and len(data['value']) > 10
+            if payment_method is not None:
+                assert type(payment_method) is str
         except AssertionError:
             raise UserError(400, 'Bad request')
+
+        if payment_method is not None:
+            PaymentMethod._api_retrieve(payment_method)  # to return 404 if not existant
 
         # All exceptions must be raised before this point.
         super().__init__()
@@ -771,6 +776,9 @@ class Customer(StripeObject):
         self.discount = None
         self.shipping = None
         self.default_source = None
+
+        if payment_method is not None:
+            PaymentMethod._api_attach(payment_method, customer=self.id)
 
         self.sources = List('/v1/customers/' + self.id + '/sources')
         self.tax_ids = List('/v1/customers/' + self.id + '/tax_ids')
--- a/test.sh
+++ b/test.sh
@@ -866,3 +866,36 @@ txn=$(
        -d charge=$charge \
   | grep -oE 'txn_\w+')
 [ -n "$txn" ]
+
+# creating a customer with a payment method
+pm=$(curl -sSfg -u $SK: $HOST/v1/payment_methods \
+          -d type=card \
+          -d card[number]=4242424242424242 \
+          -d card[exp_month]=12 \
+          -d card[exp_year]=2020 \
+          -d card[cvc]=123 \
+    | grep -oE 'pm_\w+' | head -n 1)
+
+cus=$(curl -sSfg -u $SK: $HOST/v1/customers \
+           -d description='This customer will have a payment_method when created' \
+           -d payment_method=$pm \
+     | grep -oE 'cus_\w+' | head -n 1)
+
+card=$(curl -sSfg -u $SK: $HOST/v1/payment_methods?customer=$cus\&type=card \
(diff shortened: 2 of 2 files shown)


meshy:
I wasn't sure if delegating to `PaymentMethod._api_attach` was the optimal approach. Please let me know if there's anything I can do to improve this!

localstripe/resources.py
```diff
@@ -772,6 +774,9 @@ def __init__(self, name=None, description=None, email=None,
         self.shipping = None
         self.default_source = None
 
+        if payment_method:
```
bravier, line 777: Can you check for `payment_method is not None` here too?

localstripe/resources.py
```diff
@@ -738,6 +738,8 @@ def __init__(self, name=None, description=None, email=None,
                 assert type(invoice_settings['default_payment_method']) is str
                 assert (invoice_settings['default_payment_method']
                         .startswith('pm_'))
+            if payment_method is not None:
+                assert type(payment_method) is str
```
bravier, line 742: Can you move this block just before `except AssertionError:`?
We keep the different properties checks ordered as they are passed in `__ini__`.
Thanks!

bravier (commented):
Hi @meshy!
Thanks for this contribution :+1:
Just a few comments but it looks good to me.

localstripe/resources.py
meshy: Addressed in c62a417

localstripe/resources.py
meshy: Addressed in a02f078

meshy:
> Hi @meshy!
> Thanks for this contribution +1
> Just a few comments but it looks good to me.

It's a pleasure -- thanks for this great project!

Comments addressed!

bravier:
Nice!
Can you squash the commits into one?

meshy:
> Can you squash the commits into one?

Done!

localstripe/resources.py
```diff
@@ -772,6 +774,9 @@ def __init__(self, name=None, description=None, email=None,
         self.shipping = None
         self.default_source = None
 
+        if payment_method is not None:
+            PaymentMethod._api_attach(payment_method, customer=self.id)
```
bravier, line 778: Just a final request: can you check the `payment_method` we try to attach really exists before?
Something like we do here for the `invoice` - just before the `super().__init__()`:
https://github.com/adrienverge/localstripe/blob/3a13cf8301ef43a25377e9c16b3d959577207999/localstripe/resources.py#L1564-L1565

meshy: I have found that the endpoint already seems to return a 404 when provided with an invalid `payment_method`:
```bash
# creating a customer with a non-existant payment_method returns a 404                                
code=$(curl -sg -o /dev/null -w '%{http_code}' -u $SK: $HOST/v1/customers \                           
            -d description='This customer will have a payment_method when created' \                   
            -d payment_method='pm_doesnotexist')                                                       

[ "$code" -eq 404 ]  
```

Would you like me to add this test?

Does it still seem worth adding the check?

bravier: > I have found that the endpoint already seems to return a 404 when provided with an invalid payment_method

Yes correst, `_api_attach()` returns 404 but the goal of adding this check is to avoid creating the customer if the payment method provided doesn't exist.

meshy: Ah! That makes sense, thanks. OK, I'll get on that.

meshy: Rebased and pushed :)

bravier: Looks almost perfect. It would be better to differentiate the payment method check which returns 404 and the input values checks which returns 400 error.
Can you update the code so it looks like this:
```py
try:
    ...
    if payment_method is not None:
        assert type(payment_method) is str
except AssertionError:

if payment_method is not None:
    PaymentMethod._api_retrieve(payment_method)  # to return 404 if not existant

# All exceptions must be raised before this point.
super().__init__()

self.name = name
...

if payment_method is not None:
    PaymentMethod._api_attach(payment_method, customer=self.id)
```
The test you added is nice!

localstripe/resources.py
```diff
@@ -751,6 +751,10 @@ def __init__(self, name=None, description=None, email=None,
                 assert set(data.keys()) == {'type', 'value'}
                 assert data['type'] in ('eu_vat', 'nz_gst', 'au_abn')
                 assert type(data['value']) is str and len(data['value']) > 10
+            if payment_method is not None:
+                assert type(payment_method) is str
+                PaymentMethod._api_retrieve(payment_method)
+
```
bravier, line 757: And also this blank line can be removed.

meshy: oops!

localstripe/resources.py
meshy: > It would be better to differentiate the payment method check which returns 404 and the input values checks which returns 400 error.

Good call. Done. 

> The test you added is nice!

Thanks!

localstripe/resources.py
meshy: Done

bravier (approved):
LGTM!
Thanks for your first contribution :+1:

meshy:
Thanks for the quick turn-around! :slightly_smiling_face:

adrienverge (approved):
Thanks @meshy :+1:

PR #184, opened 2021-06-28
Make tax rates properties default to empty array when unset

bravier:
**InvoiceItem: Make tax_rates property default to an empty array**

Real Stripe servers return an empty array if there is no tax rates on an
InvoiceItem object. Let's do that in localstripe too.

https://stripe.com/docs/api/invoiceitems/object#invoiceitem_object-tax_rates

---

**SubscriptionItem: Make tax_rates property default to an empty array**

Real Stripe servers return an empty array if there is no tax rates on a
SubscriptionItem object. Let's do that in localstripe too.

https://stripe.com/docs/api/subscription_items/object#subscription_item_object-tax_rates

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1583,7 +1583,7 @@ class InvoiceItem(StripeObject):
         self.period = dict(start=period_start, end=period_end)
         self.proration = proration
         self.description = description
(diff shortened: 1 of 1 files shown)


adrienverge (approved):
Nice, thanks :+1:

PR #193, opened 2021-09-03
Add bare bones support for Price

t3hmrman:
Add barebones support for price and some other changes


t3hmrman:
Apologies, a bunch of other changes are in here -- will cut it down to just the one commit with Price support

PR #195, opened 2021-09-06
feat(usage): add usage records anf fix sub item error

pandomic:
* Fix an error when listing subscription items
* Add basic implementation for usage records
* Sync latest changes on LS side

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ __pycache__
 *.py[cod]
 /dist
 /*.egg-info
+
+.idea
+.venv
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM python:3.8-slim
+
+RUN mkdir /app
+WORKDIR /app
+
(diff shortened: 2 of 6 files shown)


H--o-l:
Hey @pandomic,

Thanks for opening a PR!

A few remarks, however.

There is a lot of commits in the branch you propose, and some look outdated. 
Maybe you can rebase and keep only the commits you want this PR to be about?

My second remark is that some of your commits seem to do two things at a time, please split commits per feature or resource or parameters, the level that makes the more sense, and write a commit message as is both simple and enough to understand your goal.
Last things, commits like `fix(linter): Fix linter errors` which fix things of a previous commit of the same PR should be squashed in the previous commit.

Thanks in advance, cheers!
Hoel

pandomic:
@H--o-l thanks for your comments, we cleaned the history and opened a new PR, looking forward: https://github.com/adrienverge/localstripe/pull/197

PR #197, opened 2021-09-07
Usage records and other fixes

pandomic:
## Summary

* Fix "list" endpoint for subscription items (b24fcec)
* Add usage record endpoints (e251751)
* Add Dockerfile to easily build and push images (779726a)
* Add virtualenv's .venv and .idea folders to gitignore (b23de15)
* Frontend fix - do not throw an exception if already mounted (3b62a25)

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1651,7 +1651,7 @@ class InvoiceLineItem(StripeObject):
 
         if self.type == 'subscription':
             self.subscription_item = item.id
-            self.subscription = item._subscription
+            self.subscription = item.subscription
             self.plan = item.plan
             self.proration = False
             self.currency = item.plan.currency
@@ -3082,12 +3082,11 @@ class SubscriptionItem(StripeObject):
         self.quantity = quantity
         self.tax_rates = tax_rates
         self.metadata = metadata or {}
-
-        self._subscription = subscription
+        self.subscription = subscription
 
     def _current_period(self):
-        if self._subscription:
-            obj = Subscription._api_retrieve(self._subscription).start_date
+        if self.subscription:
+            obj = Subscription._api_retrieve(self.subscription).start_date
             start_date = obj
         else:
             start_date = int(time.time())
@@ -3143,6 +3142,26 @@ class SubscriptionItem(StripeObject):
     def _calculate_amount_in_tier(self, quantity, index):
         t = self.plan.tiers[index]
         return int(t['unit_amount']) * quantity + int(t['flat_amount'])
+
+    @classmethod
+    def _api_list_all(cls, url, subscription=None, limit=None,
+                      starting_after=None, ending_before=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        try:
+            assert _type(subscription) is str
+            assert subscription.startswith('sub_')
+            Subscription._api_retrieve(subscription)
+        except AssertionError:
+            raise UserError(400, 'Bad request')
+
+        # TODO: implement ending_before
+        li = List(url, limit=limit, starting_after=starting_after)
+        li._list = [value for key, value in store.items()
+                    if key.startswith(cls.object + ':')
+                    and value.subscription == subscription]
+        return li
 
 
 class TaxId(StripeObject):
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -3272,3 +3272,66 @@ class Token(StripeObject):
 
         self.type = 'card'
         self.card = card_obj
+
+
+class UsageRecord(StripeObject):
+    object = 'usage_record'
+    _id_prefix = 'mbur_'
+
+    class _UsageRecordSummaryWrapper(dict):
+        def _export(self, expand=None):
+            return self
+
+    def __init__(self, subscription_item_id=None, quantity=None,
+                 timestamp=None):
+        self.quantity = quantity
+        self.subscription_item = subscription_item_id
+        self.timestamp = timestamp
+
+        super().__init__()
+
+    @classmethod
+    def _api_list_summaries(cls, subscription_item_id=None, limit=None,
+                            starting_after=None, ending_before=None):
+        subscription_item = SubscriptionItem._api_retrieve(
+            subscription_item_id
+        )
+
+        usage_records = [value for key, value in store.items()
+                         if key.startswith(cls.object + ':')
+                         and value.subscription_item == subscription_item_id]
+
+        # TODO: improve summaries logic
+        summary_record = cls._UsageRecordSummaryWrapper({
+            "id": subscription_item_id.replace("si_", "sis_"),
+            "object": "usage_record_summary",
+            "invoice": subscription_item.invoice,
+            "livemode": subscription_item.livemode,
+            "period": {
+                "end": ending_before,
+                "start": starting_after
+            },
+            "subscription_item": subscription_item_id,
+            "total_usage": sum([int(r.quantity) for r in usage_records])
+        })
+
+        url = "/v1/subscription_items/%s/usage_record_summaries" \
+              % subscription_item_id
+        li = List(url, limit=limit, starting_after=starting_after)
+        li._list = [summary_record]
+
+        return li
+
+
+extra_apis.extend((
+    (
+        'POST',
+        '/v1/subscription_items/{subscription_item_id}/usage_records',
+        UsageRecord._api_create
+    ),
+    (
+        'GET',
+        '/v1/subscription_items/{subscription_item_id}/usage_record_summaries',
+        UsageRecord._api_list_summaries
+    )
+))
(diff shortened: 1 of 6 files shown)


H--o-l:
Hey @pandomic!

About the commit `fix(javascript): Be less pedantic about already modified DOM` can you describe why you need it? What is your use case? What issue does it solve for you?
Also the commit change two lines that I'm not sure are related. With an explanation inside the commit message, it would be easier for me to understand what you are doing.

About `chore(gitignore): Add .venv and .idea to gitignore` and `chore(Dockerfile): Add Dockerfile
`, localstripe doesn't need that, it's specific to your environment, and we don't want to maintain it. Can you remove them?

About the last two commits, they look great!
We will try to review them in-deep soon.

pandomic:
Hi @H--o-l, I left only subscription items and usage records-related changes. Looking forward to getting them merged.

Going back to your question, js-fix was there because of the mounting issue we had with React but leaving it out of scope for now.

I hope in the future we will see a containerized version as well ;)

test.sh
```diff
@@ -447,6 +449,10 @@ curl -sSfg -u $SK: $HOST/v1/invoices/upcoming?customer=$cus\&subscription_items[
 
 curl -sSfg -u $SK: $HOST/v1/invoices/upcoming?customer=$cus\&subscription=$sub\&subscription_items[0][id]=si_RBrVStcKDimMnp\&subscription_items[0][plan]=basique-annuel\&subscription_proration_date=1504182686\&subscription_tax_percent=20
 
+si=$(curl -sSfg -u $SK: $HOST/v1/subscription_items?subscription=$sub \
+     | grep -oE 'si_\w+')
+[ -n "$si" ]
```
H--o-l, line 456: In this first commit, `si` is not a subscription item, it's two subscription items.
So Maybe there could be a first `res` that stores them, a check of how many items there is in `res` and then, in the second commit a `si=$(echo "$res" | head -n 1)`.
It would be more readable for future readers of this code.
What do you think?

test.sh
```diff
@@ -6,6 +6,8 @@ set -eux
 HOST=http://localhost:8420
 SK=sk_test_12345
 
+curl -X DELETE $HOST/_config/data
+
```
H--o-l, line 10: Because it's not related to a localstripe feature, but it's more a CI change, this change needs to be inside another PR.
Like that we would be able to discuss it, as long as needed, without blocking this PR.

localstripe/resources.py
```diff
@@ -3144,6 +3143,26 @@ def _calculate_amount_in_tier(self, quantity, index):
...
+    @classmethod
+    def _api_list_all(cls, url, subscription=None, limit=None,
+                      starting_after=None, ending_before=None, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        try:
+            assert _type(subscription) is str
+            assert subscription.startswith('sub_')
+            Subscription._api_retrieve(subscription)
```
H--o-l, line 3155: Great to add this line!
But can it be moved outside the `try/catch`, I think there is no reason for this line to be in it.
Also can you add a small comment like `# to return 404 if not existant` on this line?

H--o-l (commented):
Hey @pandomic !

Again thanks for your work on this PR!

Here is the first series of requests, I mostly check the first commit.

test.sh
pandomic: 👍 split into proposed steps

test.sh
pandomic: 👍 removed this line

pandomic:
Hi @H--o-l, thanks for the feedback, I added proposed changes and force-pushed them to the current branch

localstripe/resources.py
```diff
@@ -165,6 +165,7 @@ def _api_list_all(cls, url, limit=None, starting_after=None, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
+        # TODO: implement ending_before
```
H--o-l, line 168: Hey!
Please remove this TODO.

localstripe/resources.py
```diff
@@ -3144,6 +3144,30 @@ def _calculate_amount_in_tier(self, quantity, index):
         t = self.plan.tiers[index]
         return int(t['unit_amount']) * quantity + int(t['flat_amount'])
 
+    @classmethod
+    def _api_list_all(cls, url, subscription=None, limit=None,
+                      starting_after=None, ending_before=None, **kwargs):
```
H--o-l, line 3149: Please remove `ending_before=None` here, so users know it's not implemented.

localstripe/resources.py
```diff
@@ -3144,6 +3144,30 @@ def _calculate_amount_in_tier(self, quantity, index):
...
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        # to return 404 if not existant
+        Subscription._api_retrieve(subscription)
```
H--o-l, line 3154: Please move these two lines after the `try/except` like in the other interfaces.

localstripe/resources.py
```diff
@@ -3144,6 +3144,30 @@ def _calculate_amount_in_tier(self, quantity, index):
...
+            assert subscription.startswith('sub_')
+        except AssertionError:
+            raise UserError(400, 'Bad request')
+
+        li = super(SubscriptionItem, cls)._api_list_all(
+            url, limit=limit, starting_after=starting_after
+        )
+
+        li._list = [obj for obj in li._list
+                    if obj.subscription == subscription]
```
H--o-l, line 3167: A quick diff proposition here:
- don't add a new line before `)` so we save one line
- don't add a new line after `)` so we save one line, and it's more similar to the other API

```diff
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -3160,9 +3160,7 @@ class SubscriptionItem(StripeObject):
             raise UserError(400, 'Bad request')
 
         li = super(SubscriptionItem, cls)._api_list_all(
-            url, limit=limit, starting_after=starting_after
-        )
-
+            url, limit=limit, starting_after=starting_after)
         li._list = [obj for obj in li._list
                     if obj.subscription == subscription]

File: localstripe/resources.py [hunk 8]
```diff
@@ -3277,3 +3277,66 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+
+    class _UsageRecordSummaryWrapper(dict):
+        def _export(self, expand=None):
+            return self
+
+    def __init__(self, subscription_item_id=None, quantity=None,
+                 timestamp=None):
+        self.quantity = quantity
+        self.subscription_item = subscription_item_id
+        self.timestamp = timestamp
```
H--o-l, line 3294: Can you add check of `kwargs` and assert of parameters like in the other ressources `__init__` function?
Assert should follow the Stripe API documentation on what is mandatory and what's not.

localstripe/resources.py
```diff
@@ -3277,3 +3277,66 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+
+        super().__init__()
```
H--o-l, line 3296: Can you add a small `# All exceptions must be raised before this point.` above this line like in the other ressources `__init__`?

localstripe/resources.py
```diff
@@ -3277,3 +3277,66 @@ def __init__(self, card=None, customer=None, **kwargs):
...
+    @classmethod
+    def _api_list_summaries(cls, subscription_item_id=None, limit=None,
+                            starting_after=None, ending_before=None):
+        subscription_item = SubscriptionItem._api_retrieve(
+            subscription_item_id
+        )
+
+        usage_records = [value for key, value in store.items()
+                         if key.startswith(cls.object + ':')
+                         and value.subscription_item == subscription_item_id]
```
H--o-l, line 3307: When I see this operation, and the routes below (`/v1/subscription_items/{subscription_item_id}/usage_records`, `/v1/subscription_items/{subscription_item_id}/usage_record_summaries`) I think it would be simpler and more logical to have usage record stored direcly inside the corresponding subscription item.


Also I'm OK with a new object `class UsageRecord()` but I think I'm NOK with a `class UsageRecord(StripeObject)`, because usage record are never used with the standard APIs that all other `StripeObject` implement, ie `api_create`, `api_retrieve`, `api_update`, `api_delete`, `api_list_all` and associated HTTP route and method (`GET  object/id`, `POST object/id` etc).

So I think that routes `/usage_records` and `/usage_record_summaries` should be extensions of the object subscription item.

What do you think?

test.sh
```diff
@@ -457,6 +459,27 @@ count=$(echo "$res" | grep -oP '"total_count": \K([0-9]+)')
...
+        -d timestamp=$timestamp \
+  | grep -oE 'mbur_\w+')
+[ -n "$usage_record_id" ]
+
+# get usage record summaries
+total_usage=$(
+  curl -sSfg -u $SK: $HOST/v1/subscription_items/$si/usage_record_summaries \
+  | grep -oP '"total_usage": \K([0-9]+)')
+[ "$total_usage" -eq 115 ]
+
```
H--o-l, line 482: :heart:

H--o-l (commented):
Hey @pandomic!

Here is a few new comments, I let you see, and don't hesitate if you need me.

PR #200, opened 2022-03-07
Invoice: Add 'collection_method' property

bravier:
Let's add the property `collection_method` that was missing.
For the moment, it's not implemented so it's always
`charge_automatically`.

https://stripe.com/docs/api/invoices/object#invoice_object-collection_method


adrienverge (approved):
Looks good, thanks!

bravier:
Thanks!

PR #203, opened 2022-06-22
tests: Add missing commas

willbarre2:
Everywhere in the `test.sh` file, when we use the `grep` command,
we try to catch a complete line and especially with the end comma.
Like here:
https://github.com/adrienverge/localstripe/blob/7f298f2/test.sh#L260 .

Let's add missing commas.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -783,17 +783,17 @@ status=$(
 
 # list charges
 total_count=$(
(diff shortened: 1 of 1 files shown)


willbarre2:
Commit and PR titles updated.

adrienverge:
Thank you @willbarre2!

PR #206, opened 2022-08-14
Publish Dockerfile in repository

vincent-artea:
Being able to build easily the Docker image can be useful when testing code changes against automated tests relying on the Docker image of localstripe.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM python:3
+COPY setup.py setup.cfg ./
+ADD localstripe ./localstripe
+RUN python3 setup.py sdist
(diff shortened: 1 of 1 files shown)


adrienverge:
Hello, thanks but there is no need to maintain a Dockerfile here when building it is described in the README and as simple as:
```
FROM python:3
RUN pip install localstripe
CMD ["localstripe"]
```

vincent-artea:
Hi,
I believe what is described in the README will generate a Docker image from the `localstripe` library from the Python Package index. To generate a Docker image from the code actually located on the developer's machine (and potentially modified from the orginal repository), I had to use the Dockerfile I proposed in the pull-request.

PR #207, opened 2022-10-09
feat(subscriptions): add 'always_invoice' to 'proration_behavior' options for subscription updates

arnaldop:
Also adds requirements.txt to make building easier and updates README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -236,7 +236,8 @@ To quickly run localstripe from source, and reload when a file changed:
 
 .. code:: shell
 
- find -name '*.py' | entr -r python3 -m localstripe --from-scratch
+ pip3 install -r requirements.txt
+ find . -name '*.py' | entr -r python3 -m localstripe --from-scratch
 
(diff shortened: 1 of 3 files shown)


H--o-l:
Hey @arnaldop, thanks for contributing!

About the `requirements.txt` I'm not a fan because `setup.py` describe them well already. Do a `pip3 install --user -e .` and you'd be done with it.
Also, it's common practice so I'm not sure it should ends-up in the README.

Could you reduce the title length for the second commit so it fits in a GitHub line (see the `...` in the screen below)? 
2022-10-10-102928_grim

Last thing, could you add an explanation in the commit message about what this change is, what it does, links to the real Stripe API to show us where it comes from, etc.
We don't know Stripe API by heart, and we will have to maintain the change you propose in the long term, so can you advocate your change to us?

The last thing, did add a test would make sense? If not can you explain why? If yes could you add a test?

Thanks in advance!

PR #208, opened 2022-10-10
CONTRIBUTING: Remove Karma commit guideline reference

H--o-l:
We don't enforce it, and often do otherwise, thus let's remove it from CONTRIBUTING.rst


bravier (approved):

feliixx (approved):

adrienverge (approved):
True. Thanks @H--o-l!

PR #210, opened 2023-05-03
NWST-332: add cloudbuild yaml

frankyhung93:


frankyhung93:
same mistake made twice

PR #213, opened 2023-06-15
Balance & expand crash

H--o-l:
### Customer: Add balance

Here is the Stripe documentation for balance at Customer creation:

Current balance, if any, being stored on the customer. If negative, the customer
has credit to apply to their next invoice. If positive, the customer has an
amount owed that will be added to their next invoice. The balance does not
refer to any unpaid invoices; it solely takes into account amounts that have
yet to be successfully applied to any invoice. This balance is only taken
into account as invoices are finalized.

I tried to make a basic implementation of it, I hope it's close enough to
what Stripe does.

----

### API: Don't crash when expanding a null object

When an Invoice doesn't have a Charge, and we
`-d expand[]=data.charge.refunds` that Invoice, the current code crash because
it cannot loop over a null Charge to find Refunds in it.

When one of the keys in the expand path is null, let's stop the recursive
exploration without crashing.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -233,7 +233,7 @@ class StripeObject(object):
                     id = obj[k]
                     cls = StripeObject._get_class_for_id(id)
                     obj[k] = cls._api_retrieve(id)._export()
-                if path is not None:
+                if path is not None and obj[k] is not None:
                     do_expand(path, obj[k])
         try:
             for path in expand:
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -710,10 +710,11 @@ class Customer(StripeObject):
                  phone=None, address=None,
                  invoice_settings=None, business_vat_id=None,
                  preferred_locales=None, tax_id_data=None,
-                 metadata=None, payment_method=None, **kwargs):
+                 metadata=None, payment_method=None, balance=0, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
 
+        balance = try_convert_to_int(balance)
         try:
             if name is not None:
                 assert type(name) is str
@@ -753,6 +754,7 @@ class Customer(StripeObject):
                 assert type(data['value']) is str and len(data['value']) > 10
             if payment_method is not None:
                 assert type(payment_method) is str
+            assert type(balance) is int
         except AssertionError:
             raise UserError(400, 'Bad request')
 
@@ -772,7 +774,7 @@ class Customer(StripeObject):
         self.business_vat_id = business_vat_id
         self.preferred_locales = preferred_locales
         self.metadata = metadata or {}
-        self.account_balance = 0
+        self.account_balance = - balance
         self.delinquent = False
         self.discount = None
         self.shipping = None
@@ -1099,7 +1101,7 @@ class Invoice(StripeObject):
         except AssertionError:
             raise UserError(400, 'Bad request')
 
-        Customer._api_retrieve(customer)  # to return 404 if not existant
+        cus = Customer._api_retrieve(customer)
 
         if subscription is not None:
             subscription_obj = Subscription._api_retrieve(subscription)
@@ -1125,9 +1127,8 @@ class Invoice(StripeObject):
         self.collection_method = 'charge_automatically'
         self.description = description
         self.discount = None
-        self.ending_balance = 0
         self.receipt_number = None
-        self.starting_balance = 0
+        self.starting_balance = cus.account_balance
         self.statement_descriptor = None
         self.webhooks_delivered_at = self.date
         self.status_transitions = {
@@ -1163,6 +1164,9 @@ class Invoice(StripeObject):
         self._draft = True
         self._voided = False
 
+        self.ending_balance = \
+            max(0, cus.account_balance - self.subtotal - self.tax)
+
         if not simulation and not upcoming:
             if subscription is not None:
                 subscription_obj.latest_invoice = self.id
@@ -1196,7 +1200,7 @@ class Invoice(StripeObject):
 
     @property
     def total(self):
-        return self.subtotal + self.tax
+        return max(0, self.subtotal + self.tax - self.starting_balance)
 
     @property
     def amount_due(self):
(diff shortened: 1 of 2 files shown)


localstripe/resources.py
```diff
@@ -1099,7 +1103,7 @@ def __init__(self, customer=None, subscription=None, metadata=None,
         except AssertionError:
             raise UserError(400, 'Bad request')
 
-        Customer._api_retrieve(customer)  # to return 404 if not existant
+        cus = Customer._api_retrieve(customer)  # to return 404 if not existant
```
adrienverge, line 1108: Look at other `# to return 404 if not existant` comments: they're here only when the call isn't assigned to any variable :wink:

localstripe/resources.py
```diff
@@ -229,6 +229,8 @@ def do_expand(path, obj):
                     do_expand(path, i)
             else:
                 k, path = path.split('.', 1) if '.' in path else (path, None)
+                if obj[k] is None:
+                    return
```
adrienverge, line 233: It feels a bit strange to me here, instead I propose to set it in the final `if`:

```diff
-                if path is not None:
+                if path is not None and obj[k] is not None:
                     do_expand(path, obj[k])
```

adrienverge (commented):
Looks good!

bravier (commented):
Nice thanks!
+1 with Adrien's comments + a few notes and style comments in tests :)

localstripe/resources.py
```diff
@@ -753,6 +753,10 @@ def __init__(self, name=None, description=None, email=None,
                 assert type(data['value']) is str and len(data['value']) > 10
             if payment_method is not None:
                 assert type(payment_method) is str
+            if balance is not None:
+                assert type(balance) is str
```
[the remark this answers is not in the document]
H--o-l: No, no reason, I will remove it.

localstripe/resources.py
H--o-l: [Not always](https://github.com/adrienverge/localstripe/blob/6e62d74465d7fea87fed28b42caf245fa95711d0/localstripe/resources.py#L3093) but it sounds weird now indeed, I hesitated. I will remove it.

localstripe/resources.py
H--o-l: Nice, cleaner like that indeed, thanks!

localstripe/resources.py
```diff
@@ -1196,7 +1204,9 @@ def total_tax_amounts(self):
 
     @property
     def total(self):
-        return self.subtotal + self.tax
+        return self.subtotal + self.tax - self.starting_balance \
+            if self.subtotal + self.tax - self.starting_balance >= 0 \
+            else 0
```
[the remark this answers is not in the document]
H--o-l: :heart:

test.sh
```diff
@@ -934,3 +934,20 @@ code=$(curl -sg -o /dev/null -w '%{http_code}' -u $SK: $HOST/v1/customers \
 total_count=$( curl -sSfg -u $SK: $HOST/v1/customers \
              | grep -oE '"total_count": 9,')
 [ -n "$total_count" ]
+
+cus=$(curl -sSf -u $SK: $HOST/v1/customers \
+          -d balance='-20000000' \
+      | grep -oE 'cus_\w+' | head -n 1)
```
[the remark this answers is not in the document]
H--o-l: I will keep the grep inside the `()` but ok for the `-d`, I didn't notice they were aligned with the `-sSf` in the other calls!

test.sh
```diff
@@ -934,3 +934,20 @@ code=$(curl -sg -o /dev/null -w '%{http_code}' -u $SK: $HOST/v1/customers \
...
+
+amount=$(curl -sSf -u $SK: $HOST/v1/subscriptions \
+             -d customer=$cus \
+             -d items[0][plan]=basique-mensuel \
+             -d expand[]=latest_invoice \
+         | grep -oP 'amount_due": \K([0-9]+)')
```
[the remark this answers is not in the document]
H--o-l: No idea for old commit `c7b5782` but let's do like the other requests you are right and add the `-g`

test.sh
```diff
@@ -934,3 +934,20 @@ code=$(curl -sg -o /dev/null -w '%{http_code}' -u $SK: $HOST/v1/customers \
...
+[ "$amount" -eq 0 ]
+
+charge=$(curl -sSfG -u $SK: $HOST/v1/invoices \
```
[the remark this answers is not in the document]
H--o-l: Now that I discovered the `-G + -d` syntax I find it cool but no strong opinion. 

I will add the `-g` indeed.

H--o-l:
@adrienverge @bravier PR updated following your reviews

localstripe/resources.py
```diff
@@ -753,6 +753,9 @@ def __init__(self, name=None, description=None, email=None,
                 assert type(data['value']) is str and len(data['value']) > 10
             if payment_method is not None:
                 assert type(payment_method) is str
+            if balance is not None:
+                balance = -try_convert_to_int(balance)
```
adrienverge, line 757: When looking at the surrounding lines, and also at the rest of the file, `try_convert_to_int()` is usually not used like this, and the default value `0` neither.

Can you change it to:
\- set the default value `balance=0` in the function signature,
\- set `balance = try_convert_to_int(balance)` before assertions,
\- inside check its type, without `if balance is not None:`,
\- set the negative inside `self.account_balance = -balance`?

adrienverge (commented):
One remark for code consistency, otherwise good to me :+1:

localstripe/resources.py
adrienverge: As you wish, as long as the linter agrees :wink: 

(I see we already use both in our code base.)

localstripe/resources.py
```diff
@@ -753,6 +754,7 @@ def __init__(self, name=None, description=None, email=None,
                 assert type(data['value']) is str and len(data['value']) > 10
             if payment_method is not None:
                 assert type(payment_method) is str
+            assert type(balance) is int
```
adrienverge, line 757: Just checking: it cannot be a float, right?

H--o-l: I admit I did a very basic implementation, so I didn't check that much. I just checked that with the same steps Stripe gives me ~ the same results.

For example, the customer balance is not updated after invoicing, so it's not a really good implementation after the first invoice.

But for your specific question about the type, I just checked and the doc says it's an integer, so I think we are good for this one:
2023-06-16-104243_grim

bravier (approved):
LGTM!

localstripe/resources.py
adrienverge: :+1:

adrienverge (approved):
Excellent!

PR #215, opened 2023-06-20
Invoice: Add pending_invoice_items_behavior property

H--o-l:
This commit adds the property to the Invoice creation call to avoid crashing, 
but it doesn't implement or test its behavior.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1286,7 +1286,8 @@ class Invoice(StripeObject):
                           subscription_proration_date=None,
                           subscription_tax_percent=None,  # deprecated
                           subscription_default_tax_rates=None,
-                          subscription_trial_end=None):
+                          subscription_trial_end=None,
+                          pending_invoice_items_behavior=None):
         subscription_proration_date = \
             try_convert_to_int(subscription_proration_date)
         try:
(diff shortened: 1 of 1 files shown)


localstripe/resources.py
```diff
@@ -1858,7 +1861,7 @@ def status(self):
             return 'processing'
 
     @property
-    def charges(self):
+    def charges(self):  # deprecated
```
H--o-l, line 1864: Not related to this commit but I took advantage of it to add this small comment.

test.sh
```diff
@@ -524,7 +524,8 @@ sub=$(curl -sSfg -u $SK: $HOST/v1/subscriptions/$sub_id \
            -d items[0][id]=$sub_item_id)
 
 in=$(curl -sSfg -u $SK: $HOST/v1/invoices \
-          -d customer=$cus)
+          -d customer=$cus \
+          -d pending_invoice_items_behavior=include)
```
adrienverge, line 528: If this new option is not really implemented, I'd prefer not using it in tests (I believe that's what we do for other unimplemented options).

adrienverge (commented):
Thanks!

test.sh
H--o-l: OK, I removed it!

adrienverge (approved):

PR #216, opened 2023-07-06
TaxID: Expand API to allow retrieving or deleting a single TaxID

feliixx:
### API: Add 'object' to the response on Deletion

The goal is to be closer to Stripe API. Also, create a new DeletedObject class
that will be easier to use.

---

### TaxID: Expand API to allow retrieving or deleting a single TaxID

Creating and listing TaxID have been supported for a long time, but retrieving
or deleting a single TaxID wasn't implemented, so let's change that.

Stripe doc:
- https://stripe.com/docs/api/customer_tax_ids/retrieve
- https://stripe.com/docs/api/customer_tax_ids/delete

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -948,6 +948,29 @@ class Customer(StripeObject):
         return obj.tax_ids
 
     @classmethod
+    def _api_retrieve_tax_id(cls, id, tax_id, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        obj = cls._api_retrieve(id)
+        for txi in obj.tax_ids._list:
+            if txi.id == tax_id:
+                return txi
+        raise UserError(404, 'Customer ' + id + ' does not have a tax ID with '
+                             'ID ' + tax_id)
+
+    @classmethod
+    def _api_delete_tax_id(cls, id, tax_id, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        obj = cls._api_retrieve(id)
+        tax_id_obj = cls._api_retrieve_tax_id(id, tax_id)
+        obj.tax_ids._list.remove(tax_id_obj)
+
+        return DeletedObject(tax_id_obj.id, tax_id_obj.object)
+
+    @classmethod
     def _api_list_subscriptions(cls, id, **kwargs):
         if kwargs:
             raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
@@ -1042,8 +1065,12 @@ extra_apis.extend((
      Customer._api_update_subscription),
     # This is the old API route:
     ('POST', '/v1/customers/{id}/cards', Customer._api_add_source),
+    ('GET', '/v1/customers/{id}/tax_ids', Customer._api_list_tax_ids),
+    ('GET', '/v1/customers/{id}/tax_ids/{tax_id}',
+     Customer._api_retrieve_tax_id),
     ('POST', '/v1/customers/{id}/tax_ids', Customer._api_add_tax_id),
-    ('GET', '/v1/customers/{id}/tax_ids', Customer._api_list_tax_ids)))
+    ('DELETE', '/v1/customers/{id}/tax_ids/{tax_id}',
+     Customer._api_delete_tax_id)))
 
 
 class Event(StripeObject):
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -158,7 +158,7 @@ class StripeObject(object):
         if key not in store.keys():
             raise UserError(404, 'Not Found')
         del store[key]
-        return {'deleted': True, 'id': id}
+        return DeletedObject(id, cls.object)
 
     @classmethod
     def _api_list_all(cls, url, limit=None, starting_after=None, **kwargs):
@@ -242,6 +242,14 @@ class StripeObject(object):
             raise UserError(400, 'Bad expand %s' % e)
 
         return obj
+
+
+class DeletedObject(StripeObject):
+    deleted = True
+
+    def __init__(self, id, object):
+        self.id = id
+        self.object = object
 
 
 class Balance(object):
--- a/localstripe/server.py
+++ b/localstripe/server.py
@@ -261,6 +261,8 @@ def api_extra(func, url):
             data['source_id'] = request.match_info['source_id']
         if 'subscription_id' in request.match_info:
             data['subscription_id'] = request.match_info['subscription_id']
+        if 'tax_id' in request.match_info:
+            data['tax_id'] = request.match_info['tax_id']
         expand = data.pop('expand', None)
         return json_response(func(**data)._export(expand=expand))
     return f
--- a/localstripe/server.py
+++ b/localstripe/server.py
@@ -238,8 +238,7 @@ def api_update(cls, url):
 def api_delete(cls, url):
     def f(request):
         id = request.match_info['id']
-        ret = cls._api_delete(id)
-        return json_response(ret if isinstance(ret, dict) else ret._export())
+        return json_response(cls._api_delete(id)._export())
     return f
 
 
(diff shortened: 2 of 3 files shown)


localstripe/resources.py
```diff
@@ -1035,7 +1058,11 @@ def _api_update_subscription(cls, id, subscription_id, **data):
     # This is the old API route:
     ('POST', '/v1/customers/{id}/cards', Customer._api_add_source),
     ('POST', '/v1/customers/{id}/tax_ids', Customer._api_add_tax_id),
-    ('GET', '/v1/customers/{id}/tax_ids', Customer._api_list_tax_ids)))
+    ('GET', '/v1/customers/{id}/tax_ids', Customer._api_list_tax_ids),
+    ('GET', '/v1/customers/{id}/tax_ids/{tax_id}',
+     Customer._api_retrieve_tax_id),
+    ('DELETE', '/v1/customers/{id}/tax_ids/{tax_id}',
+     Customer._api_delete_tax_id)))
```
adrienverge, line 1065: For consistency with other routes,, I propose to keep the 2 `GET` before the `POST`.

localstripe/resources.py
```diff
@@ -939,6 +939,29 @@ def _api_list_tax_ids(cls, id, **kwargs):
...
+                             'ID ' + tax_id)
+
+    @classmethod
+    def _api_delete_tax_id(cls, id, tax_id, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        obj = cls._api_retrieve(id)
+        tax_id_obj = cls._api_retrieve_tax_id(id, tax_id)
+        obj.tax_ids._list.remove(tax_id_obj)
```
adrienverge, line 961: Could it be simpler, and safer, to delete the item from `obj.tax_ids._list` that matches the ID? Instead of retrieving the whole object?

localstripe/resources.py
```diff
@@ -939,6 +939,29 @@ def _api_list_tax_ids(cls, id, **kwargs):
...
+    def _api_retrieve_tax_id(cls, id, tax_id, **kwargs):
+        if kwargs:
+            raise UserError(400, 'Unexpected ' + ', '.join(kwargs.keys()))
+
+        obj = cls._api_retrieve(id)
+        for txi in obj.tax_ids._list:
+            if txi.id == tax_id:
+                return txi
+        raise UserError(404, 'Customer ' + id + ' does not have a TaxId with '
+                             'ID ' + tax_id)
```
adrienverge, line 952: Usually in these error message we use the English name of objects: "does not have a subscription", "no such invoice"... So I would use "does not have a tax ID" (even if it repeats the word ID), what do you think?


Stripe docs:
stripe API docs

adrienverge (commented):
Nice addition

localstripe/resources.py
feliixx: I'm not sure that it would be simpler, as we would need to:
- iterate the list 
- throw a 404 if the item is not in the list
- remove the item if `tax_id` match the id of the item
Here, the first two steps are properly handled by `_api_retrieve_tax_id`, allowing us to avoid duplicating the logic. 

I don't think that it would be safer either, as tax ID can't be modified (they can only be created or deleted), so it's not possible for a tax ID to change between `_api_retrieve_tax_id` and `obj.tax_ids._list.remove` (eg if there was a concurrent request). But maybe you had something else in mind?

Last argument: we're already doing something similar in the code: https://github.com/adrienverge/localstripe/blob/cb071d2f18fa74e1e536e9e27c0733b6c7c69b8e/localstripe/resources.py#L903-L906

localstripe/resources.py
feliixx: Ok, done.

localstripe/resources.py
feliixx: ok, replaced `does not have a TaxId...` with `does not have a tax ID...`

localstripe/resources.py
adrienverge: OK, I must admit all these are good arguments :+1: 

Let's just keep in mind that `.remove()` can be tricky in some cases:
```python
class O:
    def __init__(self, a):
        self.a = a

    def __repr__(self):
        return f' {self.a} '

>>> l = [O('a'), O('b'), O('c')]
>>> l.remove(l[1])
>>> print(l)
[ a ,  c ]

>>> l = [O('a'), O('b'), O('c')]
>>> l.remove(O('b'))
>>> print(l)
ValueError: list.remove(x): x not in list
```

localstripe/server.py
```diff
@@ -262,8 +262,12 @@ async def f(request):
             data['source_id'] = request.match_info['source_id']
         if 'subscription_id' in request.match_info:
             data['subscription_id'] = request.match_info['subscription_id']
+        if 'tax_id' in request.match_info:
+            data['tax_id'] = request.match_info['tax_id']
         expand = data.pop('expand', None)
-        return json_response(func(**data)._export(expand=expand))
+        ret = func(**data)
+        return json_response(ret if not hasattr(ret, '_export') else
+                             ret._export(expand=expand))
```
adrienverge, line 270: (sorry I missed that on first review)

If I remember correctly, the spirit of these `api_update()` / `api_delete()` / `api_extra()` etc. is to only return JSON for types that we expect. Here, it looks like you modify `api_extra()` for your use-case (return direct JSON in a new *delete* case), but not the other functions. Also, if a new function returns something else than a `dict` or `None`, it cannot be serializable by `json_response()`.

Instead of this implementation, I would rather see `_api_delete_tax_id()` return a new `DeletedObject` type (maybe it should inherit from `StripeObject`) to keep things consistent. What do you think?

adrienverge: Other (simpler) option: do like in `api_delete()` and allow returning either a `dict` or an object with `_export()` (but only those two):
https://github.com/adrienverge/localstripe/blob/cb071d2f18fa74e1e536e9e27c0733b6c7c69b8e/localstripe/server.py#L242

feliixx: Declaring a new `DeletedObject` type was my initial idea until I saw https://github.com/adrienverge/localstripe/blob/cb071d2f18fa74e1e536e9e27c0733b6c7c69b8e/localstripe/resources.py#L156-L161

:smile: (it's the only place where we're doing this though).
So I did it this way to keep the code similar, but I agree that this `ret if not hasattr(ret, '_export')` is ugly. I'll update both call site to use a `DeletedObject` if you agree!

feliixx: > Other (simpler) option: do like in api_delete() and allow returning either a dict or an object with _export() (but only those two):

Our comments crossed, +1 for this solution that's indeed simpler

adrienverge: :+1: 

Both options (`isinstance(ret, dict)` and `DeletedObject`) are good to me, with a slight preference for `DeletedObject` which seems more extendable.

localstripe/resources.py
```diff
@@ -244,6 +244,14 @@ def do_expand(path, obj):
         return obj
 
 
+class DeletedObject(StripeObject):
+    deleted = True
+
+    def __init__(self, id, object):
+        self.id = id
+        self.object = object
```
adrienverge, line 252: An optimization could make it simpler to call the class:
```suggestion
    def __init__(self, deleted_object):
        self.id = deleted_object.id
        self.object = deleted_object.object
```

(take it or leave it)

localstripe/server.py
adrienverge: Much more satisfying :slightly_smiling_face:

adrienverge (approved):
This is great, thanks for contributing!

bravier (approved):
Looking good and clean :ok_hand:

H--o-l (approved):
Good for me too!

localstripe/resources.py
feliixx: we don't have the `deleted_object` in `_api_delete`, we only have its `id` and its class, so doing this would require to fetch the object first like this:

```python
deleted_obj = store.get(cls.object + ':' + id)
...
del store[key]
return DeletedObject(deleted_obj)
```

which seems a bit heavy just to save an argument, so I prefer keeping it as is!

adrienverge: You're right :+1:

feliixx:
Thanks for the reviews!

PR #217, opened 2023-08-28
API: Allow serving localstripe API from a custom URL

adrienverge:
It can be desirable to use localstripe from a user-defined URL with a custom path, e.g.:

    https://domain/subpath/js.stripe.com/v3/
    https://domain/subpath/v1/payment_intents/
    https://domain/subpath/v1/customers/cus_S5TdpWqIoiuKlq/sources

instead of:

    http://domain:8420/js.stripe.com/v3/
    http://domain:8420/v1/payment_intents/
    http://domain:8420/v1/customers/cus_S5TdpWqIoiuKlq/sources

This can be achieved by using a reverse-proxy for example, but the JavaScript part of localstripe didn't support it.

This commit fixes that by dynamically computing the base API URL from which the JavaScript code is loaded.

### DIFF
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -15,11 +15,15 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-// First, get the domain from which this script is pulled:
-const LOCALSTRIPE_SOURCE = (function () {
+// First, get the URL base from which this script is pulled:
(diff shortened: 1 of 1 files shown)


bravier (approved):
Thanks!

feliixx (approved):
Nice, thanks for tackling this!

adrienverge:
Thanks for reviewing!

PR #224, opened 2024-09-03
Support preauth and refunds on PaymentIntents

bpcreech:
localstripe already supports Charges with pre-auths and refunds, and it supports PaymentIntents without pre-auths or refunds.

Since PaymentIntents are either preferred or supported for most APIs now, let's support pre-auths on PaymentIntents! Since the PaymentIntent implementation in localstripe just proxies most stuff onto Charge objects, we can proxy pre-auths onto them too. Same for refunds.

We are already using this at Via (ridewithvia.com) to test our Stripe integration.

The tests are a big copy-paste-modify on the Charges tests, which seem pretty comprehensive... and we add tests for refunds on both Charges and PaymentIntents.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -589,25 +589,27 @@ class Charge(StripeObject):
             raise UserError(400, 'Bad request')
 
         obj = cls._api_retrieve(id)
+        obj._capture(amount)
+        return obj
 
+    def _capture(self, amount):
         if amount is None:
-            amount = obj.amount
+            amount = self.amount
 
         amount = try_convert_to_int(amount)
         try:
-            assert type(amount) is int and 0 <= amount <= obj.amount
-            assert obj.captured is False
+            assert type(amount) is int and 0 <= amount <= self.amount
+            assert self.captured is False
         except AssertionError:
             raise UserError(400, 'Bad request')
 
         def on_success():
-            obj.captured = True
-            if amount < obj.amount:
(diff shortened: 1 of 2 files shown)


adrienverge:
Hello @bpcreech,

Thanks for contributing upstream! I have to point out that this is a high-quality pull request with clean code, explanation and tests.
(By the way can you include the message from your PR inside the commit message?)

The refactor of `Charge` looks clean, and the additions to `PaymentIntent` and `Refund` seem to make sense. I haven't tested them thoroughly, but my private test suite passes on it :ok_hand:

Do you see any points of attention we should have before merging this?

bpcreech:
Thanks @adrienverge, and thanks very much for creating and open-sourcing localstripe! It let us retire a far less confidence-building mock we had for our Stripe integration. :)

I updated the git commit message to match the PR.

I don't know of thing to look into. We're using this with a pretty extensive test suite (717 total tests!) which runs the very same tests against the Stripe test API, so I have some confidence that it is an accurate simulation *of the surface area we use*. (That said, there's lots more to the API and surely areas we aren't simulating right yet.)

The part where we (continue to) proxy to the Charges implementation seems like it *might* be regrettable some day but I don't understand enough of the nuance to know how an implementation specialized for PaymentIntents would be any different.

adrienverge:
OK :+1: 

Thanks a lot @bpcreech! Let's release it.

PR #225, opened 2024-09-17
localstripe.js: Fix LOCALSTRIPE_BASE_API on dynamical load

adrienverge:
This continues commit c89bb15 "API: Allow serving localstripe API from a custom URL" by simplifying the way to get localstripe server URL, and also fixes the case when `localstripe.js` is loaded dynamically (in this case, `document.getElementsByTagName('script')` references the wrong element: the calling/parent `<script>`) like this:

    let script = document.createElement('script');
    script.type = 'text/javascript';
    script.src = 'http://localhost:8420/js.stripe.com/v3/';
    document.head.appendChild(script);

### DIFF
--- a/localstripe/localstripe-v3.js
+++ b/localstripe/localstripe-v3.js
@@ -17,8 +17,7 @@
 
 // First, get the URL base from which this script is pulled:
 const LOCALSTRIPE_BASE_API = (function () {
-  const scripts = document.getElementsByTagName('script');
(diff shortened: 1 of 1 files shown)


feliixx (approved):
Nice :+1:

GuillaumeCz (approved):
:+1:

H--o-l (approved):
:+1:

adrienverge:
Thanks!

PR #227, opened 2024-09-25
TaxID: Add support for Spanish CIF

feliixx:
### TaxID: Correctly set country for non 'eu_vat' taxIDs

'au_abn' and 'nz_gst' are numbers with only digits, they don't contain their
country code.
We can only guess the country from 'eu_vat' taxIDs, so let's adapt the code to
better handle AU and NZ taxIDs number.

See [the Stripe documentation] for the exact format of these two taxIDs number.

[the Stripe documentation]: https://docs.stripe.com/billing/customer/tax-ids

---

### TaxID: Add support for Spanish CIF

According to Stripe's documentation[^1], two Spanish Customer Tax IDs (`eu_vat`
& `es_cif`) are supported.

[^1]: https://docs.stripe.com/billing/customer/tax-ids#supported-tax-id-types

Let's add the support for Spain's _código de identificación fiscal_ (CIF) tax
ID.

Even if Stripe doesn't validate this type of Customer Tax ID (unlike EU
VAT)[^2], the related dashboard form input only checks if the given value's
length is 9.

[^2]: https://docs.stripe.com/billing/customer/tax-ids#validation

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -3276,7 +3276,13 @@ class TaxId(StripeObject):
             assert type in ('eu_vat', 'nz_gst', 'au_abn')
             assert _type(value) is str and len(value) > 10
             if country is None:
-                country = value[0:2]
+                if type == 'eu_vat':
+                    country = value[0:2]
+                elif type in ('nz_gst', 'au_abn'):
+                    country = type[0:2].upper()
+                else:
+                    # shouldn't happen because type is checked above
+                    assert False
             assert _type(country) is str
         except AssertionError:
             raise UserError(400, 'Bad request')
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -759,8 +759,12 @@ class Customer(StripeObject):
             for data in tax_id_data:
                 assert type(data) is dict
                 assert set(data.keys()) == {'type', 'value'}
-                assert data['type'] in ('eu_vat', 'nz_gst', 'au_abn')
-                assert type(data['value']) is str and len(data['value']) > 10
+                assert data['type'] in ('eu_vat', 'nz_gst', 'au_abn', 'es_cif')
+                assert type(data['value']) is str
+                if data['type'] == 'es_cif':
+                    assert len(data['value']) == 9
+                else:
+                    assert len(data['value']) > 10
             if payment_method is not None:
                 assert type(payment_method) is str
             assert type(balance) is int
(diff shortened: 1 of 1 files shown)


localstripe/resources.py
```diff
@@ -3276,7 +3276,10 @@ def __init__(self, country=None, customer=None, type=None, value=None,
             assert type in ('eu_vat', 'nz_gst', 'au_abn')
             assert _type(value) is str and len(value) > 10
             if country is None:
-                country = value[0:2]
+                if type == 'eu_vat':
+                    country = value[0:2]
+                else:
+                    country = type[0:2].upper()
```
adrienverge, line 3294: Suggestion:
```suggestion
                if type == 'eu_vat':
                    country = value[0:2]
                elif type in ('nz_gst', 'au_abn'):
                    country = type[0:2].upper()
                else:
                    raise
```

The goal is to avoid "missing" cases if one day we add another value in the list `('eu_vat', 'nz_gst', 'au_abn')`.

(and `nz_gst, au_abn` → `nz_gst, au_abn, es_cif` in the next commit)

adrienverge (commented):
Nice addition! Thanks

localstripe/resources.py
feliixx: Ok interesting ! Do you really suggest adding a naked return for "missing" case (I thought this was only for re-raising an exception in an `except` block), or do you mean something like `raise UserError(500, 'Not implemented')` (that we're already using elsewhere in the codebase) ?

H--o-l (commented):
Interesting!

One question on the first commit, and the second one looks :ok_hand:

localstripe/resources.py
```diff
@@ -3276,7 +3276,10 @@ def __init__(self, country=None, customer=None, type=None, value=None,
             assert type in ('eu_vat', 'nz_gst', 'au_abn')
             assert _type(value) is str and len(value) > 10
             if country is None:
-                country = value[0:2]
```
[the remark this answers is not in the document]
feliixx: sample `eu_vat`: `ATU12345678`
sample `au_abn`: `12345678912`
sample `nz_gst`: `123456789`

(cf https://docs.stripe.com/billing/customer/tax-ids) 

We can use the first two characters of the **value** is the country code **only** for `eu_vat`. For the two others taxIDs, it's the first two characters of the **type** (and it seems nice to use this), hence the if/else.

localstripe/resources.py
adrienverge: You're right, `raise` is a bit brutal, and there is no exception to reraise...

`raise UserError` wouldn't be suited, because it's not a user error, it would be a bug on our side (because we already checked `type` is amongst valid values a few lines above). We would need something like `raise ABugWasIntroducedInOurCodeAndNeedsToBeFixed()`. Since this doesn't exist (yet) in localstripe, I propose `assert False  # shouldn't happen because type is checked above`. What do you think?

feliixx: >  I propose assert False  # shouldn't happen because type is checked above. 

Ok, done in last push.

localstripe/resources.py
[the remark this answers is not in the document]
H--o-l: Of course, my bad, I didn't read the code carefully enough.

Thanks for the explanation!

H--o-l (approved):
OK for me too :ok_hand:

PR #229, opened 2024-09-25
Add back failure if no payment method on confirm

bpcreech:
I added the ability to add a payment method by ID on confirm here: https://github.com/adrienverge/localstripe/pull/226

I removed validation that there was payment_method_data.

Per review comment on that PR, common sense, and testing with the actual Stripe API, we should still return a 400 error if neither payment_method nor payment_method_data are supplied to the confirm request. So I added that and changed an old test case that seemed to be asserting a behavior unlike what Stripe actually does.

This also addresses another review comment, I was resetting payment_method_types within the confirm handler. Upon testing, this is not what the real Stripe API does, so I removed that recent addition.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -2769,16 +2769,18 @@ class SetupIntent(StripeObject):
             pm = PaymentMethod(**payment_method_data)
             obj._attach_pm(pm)
         elif obj.payment_method is None:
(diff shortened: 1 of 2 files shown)


bpcreech:
Oh, wow, I realize this is both correct and the total opposite of what the original comment was asking about.

Let me take this out of review and try again.

PR #230, opened 2024-09-25
Fail if no payment method on confirm

bpcreech:
I added the ability to add a payment method by ID on confirm here: https://github.com/adrienverge/localstripe/pull/226

Per testing with the actual Stripe API, and common sense, we should return a 400 error if neither payment_method nor payment_method_data are supplied to the confirm request (the point of the confirm is to add a payment method, so it makes no sense to call confirm with no payment method). So I added validation and changed an old test case that seemed to be asserting a behavior unlike what Stripe actually does.

This also addresses a one-liner review comment from this PR: https://github.com/adrienverge/localstripe/pull/226

I was resetting payment_method_types within the confirm handler. Upon testing, this is not what the real Stripe API does, so I removed that recent addition.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -2769,16 +2769,18 @@ class SetupIntent(StripeObject):
             pm = PaymentMethod(**payment_method_data)
             obj._attach_pm(pm)
         elif obj.payment_method is None:
-            obj.status = 'requires_payment_method'
-            obj.next_action = None
+            # If no payment method was specified upon SetupIntent creation, and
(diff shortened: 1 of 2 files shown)


bpcreech:
this is the same as https://github.com/adrienverge/localstripe/pull/229 but with a corrected commit message. :) Now I'm doing the right thing for the right reasons, instead of the right thing for the wrong reasons!

H--o-l (approved):
OK thanks, good for me!

adrienverge (approved):
I haven't tested it, but it makes sense :+1:

PR #231, opened 2024-09-26
Fix pre-auth behavior

bpcreech:
Before this change we were erroneously marking pre-auth'd charges as status=pending, when they're actually status=succeeded. We were (accidentally) working around this incorrect behavior in pre-auth'd PaymentIntents.

To get this right we have to actually split the _trigger_payment method into two: a check for payment authorization (which we do on construction even for Charges created with capture=false), and a separate routine to actually capture the charge (which we do on construction for non-pre-auth'd charges, and on _api_capture for pre-auth'd charges). We then adjust the PaymentIntent wrapper to fit.

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -506,6 +506,7 @@ class Charge(StripeObject):
         self.status = 'pending'
         self.receipt_email = None
         self.receipt_number = None
+        self.payment_intent = None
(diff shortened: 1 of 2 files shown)


bpcreech:
oh whoops I just noticed I didn't finish the last test in this, will fix and re-open

PR #233, opened 2024-10-08
Subscription: Test that creation with card 4000000000000341 works

adrienverge:
This adds a small suite of tests related to the discussion at https://github.com/adrienverge/localstripe/pull/232#issuecomment-2400000513

It seems that Stripe works this way (I've tested it today), so let's make sure Localstripe does it too, by adding a non-regression test.


H--o-l (approved):
:ok_hand:

PR #240, opened 2024-10-25
Add support for off_session 3D Secure

bpcreech:
This adds a simulation of a canonical Stripe test card that supports 3D Secure in off_session mode, if the configured properly with a setup_intent beforehand. This card stops requesting 3DS authentication at payment time if it was set up for off-session payments.

This commit includes addition of a test-only _authenticate endpoint for setup_intents similar to the one that already exists for payment_intents. (Tests can POST to the _authenticate endpoint to simulate asynchronous interaction by the cardholder with 3D Secure challenges.)


bpcreech:
BTW in a separate thread I started working on a minimal web page form as a GET for `/v1/setup_intents/:seti/_authenticate`, similar to the page that Stripe creates for redirect-based payment methods (3DSecure cards, SEPA, PayPal, probably more...). This would give us a place to actually do the redirect manually. The form would just have two buttons, POSTing to the endpoint this PR creates, and the POST endpoint would return a 302 error redirecting to the given `redirect_url`.

I actually don't rely on there being such a page but it seems like it might be a more reasonable solution if we fill in the gap on the simulation between `confirm` and `authenticate`.

Let me know if that sounds either horrible (in which case I shall stop building it), useful, or a blocking requirement *this* PR. :)

bpcreech:
> Once again, the code looks good and documented 👍
> 
> It's interesting because we already use localstripe for 3D Secure payments in off_session mode, by pre-configuring them with setup intents beforehand. So it's supposed to be working already on master branch (al least for our particular use case :)) For the authentication confirmation/failure (done by automatic tests), we use localstripe.js ([documented here](https://github.com/adrienverge/localstripe/tree/v1.15.5?tab=readme-ov-file#integrate-with-stripe-elements), sorry because maybe it's not advertised enough). What does this commit solves compared to what we can already do?
> 

Yeah good point, I clarified this in the commit message better just now. What this special test card does is **stop** requiring authentication when we confirm a PaymentIntent, if the card was set up for off-session payments *and* the payment is off-session. This seems to be a realistic behavior (as we at Via are confirming live in production for the last couple weeks!)

For context if it's interesting: At Via we are running transit systems, like [this one in Avignon](https://www.orizo.fr/se-deplacer/nos-mobilites/transport-sur-reservation)... usually the riders of our services pay using a service like Stripe on our app... and usually the rider is off-session when the payment happens—usually but not always their phone is in their pocket or purse because they're stepping out of a van when payment actually happens! So we need to simulate the realistic 3D Secure behavior: small off-session payments are less likely to trigger 3D Secure challenges if the card was set up for it, because Stripe is able to [request an exemption](https://stripe.com/guides/strong-customer-authentication#exemptions-to-strong-customer-authentication). That's what this test card simulates for us.

> > Let me know if that sounds either horrible (in which case I shall stop building it), useful, or a blocking requirement this PR. :)
> 
> It looks like a good idea! But it might be redundant with localstripe.js's `confirmCardSetup()` and `confirmCardPayment()`? Same for the new test-only endpoint `/v1/setup_intents/{id}/_authenticate`?

... okay still looking at this!

bpcreech:
> It looks like a good idea! But it might be redundant with localstripe.js's confirmCardSetup() and confirmCardPayment()? Same for the new test-only endpoint /v1/setup_intents/{id}/_authenticate?

Oh I see... what `confirmCardSetup()` does is just poke `/v1/setup_intents/{id}/confirm` twice. The second time works.

Let me make an analogy here:

1. The "front door" for *starting a SetupIntent confirmation* is `/v1/setup_intents/{id}/confirm`.
2. The "front door" for *authenticating a SetupIntent* is a Stripe proprietary endpoint which, in the case of 3DS, you can't even get to without evaluating a bunch of JS.
3. The "back door" for *authenticating a SetupIntent* I have added here is `/v1/setup_intents/{id}/_authenticate`.
4. But there was already a "back door" for *authenticating a SetupIntent*: just call POST to `/v1/setup_intents/{id}/confirm` a second time.

So this addition is indeed redundant.

Eh, as a matter of taste, it might be a good idea to avoid intermingling item #1 ("The "front door" for *starting a SetupIntent confirmation*") with item #3 (The "back door" for *authenticating a SetupIntent*). To do that, what I would do is modify localstripe-v3.js (in this PR) to switch to my new endpoint for the authenticate step. I don't know if that accomplishes much, just, maybe, provides future simplification of `SetupIntent._api_confirm` because it would no longer need to simutaneously act as "front door for confirmation" and "backdoor for authentication". My only hesitation is that I don't have a test setup for `localstripe-v3.js` and haven't yet figured out how to even use it; the work I'm doing is all backend.

bpcreech:
> > It looks like a good idea! But it might be redundant with localstripe.js's confirmCardSetup() and confirmCardPayment()? Same for the new test-only endpoint /v1/setup_intents/{id}/_authenticate?
> 
> Oh I see... what `confirmCardSetup()` does is just poke `/v1/setup_intents/{id}/confirm` twice. The second time works.
> 
> Let me make an analogy here:
> 
> 1. The "front door" for _starting a SetupIntent confirmation_ is `/v1/setup_intents/{id}/confirm`.
> 2. The real-life "front door" for _authenticating a SetupIntent_ is a Stripe proprietary endpoint which, in the case of 3DS, you can't even get to without evaluating a bunch of JS.
> 3. The simulated "back door" for _authenticating a SetupIntent_ I have added here is `/v1/setup_intents/{id}/_authenticate`.
> 4. But there was already a simulated "back door" for _authenticating a SetupIntent_: just call POST to `/v1/setup_intents/{id}/confirm` a second time.
> 
> So this addition is indeed redundant.
> 
> Eh, as a matter of taste, it might be a good idea to avoid intermingling item #1 ("The "front door" for _starting a SetupIntent confirmation_") with item #3 (The "back door" for _authenticating a SetupIntent_). To do that, what I would do is modify localstripe-v3.js (in this PR) to switch to my new endpoint for the authenticate step. I don't know if that accomplishes much, just, maybe, provides future simplification of `SetupIntent._api_confirm` because it would no longer need to simutaneously act as "front door for confirmation" and "backdoor for authentication". My only hesitation is that I don't have a test setup for `localstripe-v3.js` and haven't yet figured out how to even use it; the work I'm doing is all backend.

PR #241, opened 2024-10-25
Prevent refunds on a failed charge

bpcreech:


adrienverge (approved):
Of course, makes sense!

Thanks again

PR #244, opened 2025-01-27
Invoice: Add `customer_email` in invoice object

feliixx:
[Stripe documentation on invoices] describes this field like this:              
                                                                                
> customer_email - nullable string                                              
>                                                                               
> The customer’s email. Until the invoice is finalized, this field will equal   
> customer.email. Once the invoice is finalized, this field will no longer be   
> updated.                                                                      
                                                                                
This is the field that should be used to know where an invoice was effectively  
sent (when using Stripe or a third-party solution to send invoices by email)    
                                                                                
Note that the current implementation only updates the field on invoice          
finalization to keep the implementation simple.                                 
                                                                                
[Stripe documentation on invoices]:                                             
https://docs.stripe.com/api/invoices/object#invoice_object-customer_email

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1208,6 +1208,7 @@ class Invoice(StripeObject):
         super().__init__()
 
         self.customer = customer
+        self.customer_email = cus.email
         self.subscription = subscription
         self.tax_percent = tax_percent
         self.default_tax_rates = default_tax_rates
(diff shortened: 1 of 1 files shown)


localstripe/resources.py
```diff
@@ -1334,6 +1335,8 @@ def charge(self):
     def _finalize(self):
         assert self.status == 'draft'
         self._draft = False
+        cus = Customer._api_retrieve(self.customer)
+        self.customer_email = cus.email
```
adrienverge, line 1339: Could this be simplified in one line?

adrienverge (approved):
Looks good

H--o-l (approved):
:+1:

localstripe/resources.py
feliixx: Yes, done in last push :+1:

bhubr (approved):

PR #249, opened 2025-08-29
Invoice: Add support for Create Preview Invoice API

bravier:
The Upcoming Invoice API is deprecated and removed since API version 2025-03-31:
https://docs.stripe.com/changelog/basil/2025-03-31/invoice-preview-api-deprecations

The new Create Preview Invoice API is almost the same and only the parameters structure changed.

Reference: https://docs.stripe.com/api/invoices/create_preview

### DIFF
--- a/localstripe/resources.py
+++ b/localstripe/resources.py
@@ -1603,6 +1603,40 @@ class Invoice(StripeObject):
         return invoice
 
     @classmethod
+    def _api_create_preview_invoice(cls, customer=None, subscription=None,
+                                    subscription_details=None):
+        try:
+            if subscription_details is not None:
+                assert type(subscription_details) is dict
+        except AssertionError:
+            raise UserError(400, 'Bad request')
+
+        details = subscription_details or {}
+        default_tax_rates = details.get('default_tax_rates')
+        items = details.get('items')
+        proration_date = details.get('proration_date')
+        trial_end = details.get('trial_end')
+
+        invoice = cls._get_next_invoice(
+            customer=customer,
+            subscription=subscription,
+            upcoming=True,
+            subscription_default_tax_rates=default_tax_rates,
+            subscription_items=items,
+            subscription_proration_date=proration_date,
+            subscription_trial_end=trial_end)
+
+        # Do not store this invoice but real Stripe server do for a limited
+        # amount of time (72 hours) which make it possible to retrieve it
+        # on route /v1/invoices/:id:
+        # https://docs.stripe.com/invoicing/preview
+        del store[cls.object + ':' + invoice.id]
+
+        invoice.id = f'upcoming_{invoice.id}'
+
+        return invoice
+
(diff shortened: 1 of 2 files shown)


localstripe/resources.py
```diff
@@ -1602,6 +1602,37 @@ def _api_upcoming_invoice(cls, customer=None, subscription=None,
...
+            upcoming=True,
+            subscription_default_tax_rates=default_tax_rates,
+            subscription_items=items,
+            subscription_proration_date=proration_date,
+            subscription_trial_end=trial_end)
+
+        # Do not store this invoice
+        del store[cls.object + ':' + invoice.id]
+
+        invoice.id = f'upcoming_{invoice.id}'
```
bravier, line 1632: The Create Preview Invoice API adds an ID to the returned invoice whereas the Upcoming Invoice API didn't do that.

localstripe/resources.py
```diff
@@ -1602,6 +1602,37 @@ def _api_upcoming_invoice(cls, customer=None, subscription=None,
...
+            customer=customer,
+            subscription=subscription,
+            upcoming=True,
+            subscription_default_tax_rates=default_tax_rates,
+            subscription_items=items,
+            subscription_proration_date=proration_date,
+            subscription_trial_end=trial_end)
+
+        # Do not store this invoice
+        del store[cls.object + ':' + invoice.id]
```
feliixx, line 1630: The Stripe docs says that an `upcoming` invoice can be retrieved via the `/v1/invoices/id` for 72 hours (cf https://docs.stripe.com/invoicing/preview), so maybe it would make more sense to keep these `upcoming` invoice in the store ?

feliixx (commented):
Nice and clean !

localstripe/resources.py
adrienverge: It would work for me. I'm curious about what @feliixx thinks!

feliixx: > I don't want to implement such "temporary" store in localstripe that would complicate things.

+1 not to implement a "temporary" store: we don't need to access the preview invoices via the `/v1/invoices/:id` route right now so it's not worth the extra complexity

> Also, please note that the /v1/invoices/:id route is not supported in localstripe right now so storing the invoice would not be useful.

Ok interesting, I wasn't aware of that ! 

> Maybe I can change the comment to says that real Stripe servers keep these invoices for a limited amount of time? What do you think?

Seems like the best solution :+1:

bravier: Comment updated: [diff](https://github.com/adrienverge/localstripe/compare/019a82fa3d339d39a2b2af4215649445ecf02752..9e0c094f505ce04149ceca0d9c816b9d43c03497)

feliixx (approved):
Ok for me, thanks 👍

bravier:
Thanks!
Let's merge and release a new version.

